PR #56, opened 2022-08-13
feat: add toggle rule

tlambert03:
makes it easier to provide callbacks to initialize a Toggleable action... this exposes the backend, so is marked as experimental.  Since we likely will only have a single backend for the foreseeable future, it's a convenient and perhaps necessary workaround for now, but should be reconsidered and removed eventually

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
 ]
 dynamic = ["version"]
-dependencies = ['psygnal', 'pydantic', 'in-n-out', 'typing_extensions']
+dependencies = ['psygnal', 'pydantic', 'in-n-out>=0.1.4', 'typing_extensions']
 
 # extras
 # https://peps.python.org/pep-0621/#dependencies-optional-dependencies
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
 ]
 dynamic = ["version"]
-dependencies = ['psygnal', 'pydantic', 'in-n-out>=0.1.4', 'typing_extensions']
+dependencies = ['psygnal', 'pydantic', 'in-n-out>=0.1.5', 'typing_extensions']
 
 # extras
 # https://peps.python.org/pep-0621/#dependencies-optional-dependencies
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
 ]
 dynamic = ["version"]
-dependencies = ['psygnal', 'pydantic', 'in-n-out', 'typing_extensions>=4.0']
+dependencies = ['psygnal', 'pydantic', 'in-n-out', 'typing_extensions']
 
 # extras
 # https://peps.python.org/pep-0621/#dependencies-optional-dependencies
@@ -54,6 +54,7 @@ docs = [
     "mkdocstrings-python==0.7.0",
     "mkdocstrings==0.19.0",
     "mkdocs-macros-plugin==0.7.0",
+    "typing_extensions>=4.0",
 ]
 [project.urls]
 homepage = "https://github.com/napari/app-model"
@@ -118,10 +119,7 @@ ignore = "D100,D213,D401,D413,D107"
 # https://docs.pytest.org/en/6.2.x/customize.html
 [tool.pytest.ini_options]
 minversion = "6.0"
-filterwarnings = [
-    "error",
-    "ignore:QPixmapCache.find:DeprecationWarning:",
-]
+filterwarnings = ["error", "ignore:QPixmapCache.find:DeprecationWarning:"]
 
 # https://mypy.readthedocs.io/en/stable/config_file.html
 [tool.mypy]
--- a/src/app_model/types/_command_rule.py
+++ b/src/app_model/types/_command_rule.py
@@ -24,8 +24,9 @@ class ToggleRule(_StrictModel):
         None,
         description="Experimental! Function to be called when a menu item/button is "
         "first created. Must take a single argument (the backend object, e.g. a "
-        "`QAction`) and may perform signal connection. Since this exposes the backend, "
-        "this field may be removed in the future.",
+        "`QAction` instance) and may perform signal connection. Since this exposes "
+        "the backend, this field may be removed in the future. Prefer using `condition`"
+        " instead.",
     )
 
 
--- a/src/app_model/types/_command_rule.py
+++ b/src/app_model/types/_command_rule.py
@@ -1,10 +1,32 @@
-from typing import Optional
+from typing import Any, Callable, Optional, Union
 
 from pydantic import Field
 
 from .. import expressions
 from ._base import _StrictModel
 from ._icon import Icon
+
+
+class ToggleRule(_StrictModel):
+    """More detailed description of a toggle rule."""
+
+    condition: Optional[expressions.Expr] = Field(
+        None,
+        description="(Optional) Condition under which the command should appear "
+        "checked/toggled in any GUI representation (like a menu or button).",
+    )
+    initialize: Optional[Callable[[], bool]] = Field(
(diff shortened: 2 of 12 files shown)


src/app_model/types/_command_rule.py
```diff
@@ -7,6 +7,29 @@
 from ._icon import Icon
 
 
+class ToggleRule(_StrictModel):
```
tlambert03, line 10: @Czaki, if you have a moment to take a look, would appreciate your thoughts.

This is my current strategy for getting the viewer toggle action to work in https://github.com/napari/napari/pull/4826

(see https://github.com/napari/napari/pull/4826/files#diff-b93fcae709b97acc12b9c786489465f3d8b8dcb186d5035c0df9a7b49b7b0e36 for the napari side usage)

the "better" long-term solution is to just use a when clause to determine whether the toggle should be checked or not.  but that would require a lot more work on the side of context names in napari.  This provides a more direct hookup (akin to what we're already doing in napari).  

The obvious downside is that it passes the backend object directly... hence the name `experimental_connect`.  I figure we'll have the context/when clauses finished long before we use a backend other than Qt ... so I (personally) feel like this workaround is fine for the short/medium term.  

thoughts?

pyproject.toml
```diff
@@ -20,7 +20,7 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
 ]
 dynamic = ["version"]
-dependencies = ['psygnal', 'pydantic', 'in-n-out', 'typing_extensions']
+dependencies = ['psygnal', 'pydantic', 'in-n-out>=0.1.5', 'typing_extensions']
```
Czaki, line 23: I do not see `typing_extensions` usage in this PR. Is this not added earlier, or is some artifact from the development phase?

src/app_model/types/_command_rule.py
Czaki: At this moment, it is not clear to me how it will work for viewer. Did condition will be passed to injection manager that will provide values to it calculate? 

Both `initialize` and `experimental_connect` are pointed to be called only on creation. So will not allow to synchronize state changed from another place.

tests/test_qt/test_qactions.py
```diff
@@ -0,0 +1,56 @@
...
+    a1.trigger()
+    assert a1.isChecked()
+    assert x
+    mock.assert_called_once_with(True)
+    mock.reset_mock()
+
+    a1.trigger()
+    assert not a1.isChecked()
+    assert not x
+    mock.assert_called_once_with(False)
```
Czaki, line 56: here I do not see checking if the change of the `x` value from the code will be reflected in the `a1` state.

pyproject.toml
tlambert03: yeah, it's not added or new in this PR,  the change in this line is to `in-n-ou`

tests/test_qt/test_qactions.py
tlambert03: correct, because app-model makes no assumption that `x` is somehow "evented".  The test you're looking for here would be made on the napari side, essentially: to check that changing the value of some evented object x calls `a1.trigger()` ... the corresponding test here would simply be to call `trigger()` directly; that is, we're not testing event systems here.

src/app_model/types/_command_rule.py
tlambert03: it's supposed to work as follows:
1. viewer.window gets created
2. menus get built, calling `build_qmodel_menu` for each of the top menus
3. `QModelMenu` then creates a `QMenuItemAction` for each action in the menu
4. during the init of the MenuItemAction, `experimental_connect()` gets called.
5. at that point, we have a current viewer, so if the callback provided to experimental connect has any annotations, they can be injected.

There _is_ however, a sneaky little bug in here that needs to be fixed; and that is that `QMenuItemAction` objects are cached (so that different representations of the same command in a menu, button, etc. all point to the same QAction)... however, that caching means that step 4/5 are only getting called for the first viewer.  So indeed, the second viewer isn't synced.  need to ponder that :/

tests/test_qt/test_qactions.py
Czaki: But there should be a such an option like `refresh()` that will evaluate `condition` to check if it should be checked. Then the connection could be made on the side of the backend without exposing this on the side of the definition.

tlambert03: but what would trigger the call of refresh?  either we poll periodically (obviously bad) or the consumer (napari) needs to tell us when something appropriate has happened right?  and that would be an event?  (sorry if I'm missing something obvious here!)

Czaki: If we talk about menus, the trigger could be on the open menu (https://doc.qt.io/qt-6/qmenu.html#aboutToShow). 

I see more scenarios, but this require things from `experimental_connect` 

Having such `refresh()` function could allow to remove `experimental_connect` and have something that could be connected to signals.

pyproject.toml
Czaki: Ahh. But still, typing_extenstions should be in version at least 3.10?

tlambert03: Well I import from it without checking python version, so yes. 

Typing extensions itself has all the version guard conditionals, so if the typing module offers the thing you're using, it will just use it.  So there's no need to only install it conditionally

Czaki: so `ParamSpec` is present in all version of `typing_extensions`? I think that 3.10.x is a minimal version of `typing_extensions` to have code working. Or I'm wrong?

src/app_model/types/_command_rule.py
Czaki: So determine state on open menu by evaluating conditions will solve this problem?

pyproject.toml
tlambert03: Oh I see what you're saying, no it's not, but as you'll see in the code usage, I specifically try to import param spec, and then fall back to using type var.  the reason here is that an old version of tensorflow very annoyingly pinned typing extensions, and so if I pin a min value here, it's incompatible with tensorflow (and we'll get an error here if someone installs TF on top).  

See https://github.com/napari/superqt/pull/101
And https://github.com/napari/app-model/pull/49

src/app_model/types/_command_rule.py
tlambert03: ah, that's definitely worth a try!  That could completely replace the need for experimental connect huh?  It would mean that any changes on the napari side wouldn't be reflected in an "open" menu... but that's such a minor concern, and will almost never happen.  so i think I like it!  will update when I try it

Czaki: The standard method of invoking the menu https://doc.qt.io/qt-6/qmenu.html#exec should prevent any changes from the main thread.

tlambert03: yeah... actually though.  I just thought of a downside of this: which is that menus are just one place where a toggleable qaction might appear.  They can also back QToolbuttons (which app-model also supports)... and in that case, when you're not opening a menu, it's less obvious when one should call "refresh".  

So, there's definitely a big benefit of using the "on menu open" approach to check toggle state, and it would mean we wouldn't have to expose the backend at all (which i love).  but it wouldn't work for qtoolbuttons: and ultimately, I think that all of our layer mode buttons should be toolbuttons.

However, that's probably even farther down the road, and at that point we could just say, "if you want a responsive toolbutton, you have to use a context condition".  Seem good?

Czaki: Hm. And what did you think about passing a tuple `(Annotation_for_injection, `path.to.signal.to.be.connected`) instead of function? this will force a single event but allow for proper connection and disconnect when creating/deleting backend objects.

tlambert03: I just made a new PR for you to review what it might look like to evaluate conditions whenever a menu is being show: https://github.com/napari/app-model/pull/59

let me have a think now about this tuple strategy...

pyproject.toml
tlambert03: gonna resolve this topic to keep the visual clutter down.  but let me know if there's anything confusing about why/what i did there

PR #59, opened 2022-08-16
feat: add toggle rule (v2)

tlambert03:
an alternative pattern to #56, suggested by @Czaki ... where the toggle rule only takes function that returns the current value.  We then call that function on the menu's `aboutToShow` signal.

This won't work for qtoolbuttons ... but that can be addressed later (and a when clause _would_ work for those buttons anyway)

closes #56

### DIFF
--- a/tests/test_qt/test_qactions.py
+++ b/tests/test_qt/test_qactions.py
@@ -1,0 +1,56 @@
+from typing import TYPE_CHECKING
+from unittest.mock import Mock
+
+from app_model.backends.qt import QCommandRuleAction, QMenuItemAction
+from app_model.types import Action, MenuItem, ToggleRule
+
+if TYPE_CHECKING:
+    from qtpy.QtWidgets import QAction
+
+    from app_model import Application
+    from conftest import FullApp
+
+
+def test_cache_qaction(qapp, full_app: "FullApp") -> None:
+    action = next(
+        i for k, items in full_app.menus for i in items if isinstance(i, MenuItem)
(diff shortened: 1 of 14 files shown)


tlambert03:
@Czaki, on the napari side, the `ViewerToggleAction` also looks simpler:

```python
class ViewerToggleAction(Action):
    def __init__(
        self,
        *,
        id: str,
        title: str,
        viewer_attribute: str,
        sub_attribute: str,
        **kwargs,
    ):
        def get_current(viewer: 'Viewer'):
            """return the current value of the viewer attribute"""
            attr = getattr(viewer, viewer_attribute)
            return getattr(attr, sub_attribute)

        def toggle(viewer: 'Viewer'):
            """toggle the viewer attribute"""
            attr = getattr(viewer, viewer_attribute)
            setattr(attr, sub_attribute, not getattr(attr, sub_attribute))

        super().__init__(
            id=id,
            title=title,
            toggled=ToggleRule(get_current=get_current),
            callback=toggle,
            **kwargs,
        )
```

tests/test_qt/test_qactions.py
```diff
@@ -0,0 +1,62 @@
...
+    assert x
+
+    a1.trigger()
+    assert not a1.isChecked()
+    assert not x
+
+    x = True
+    a1._refresh()
+    mock.assert_called_once()
+    assert a1.isChecked()
```
tlambert03, line 62: this bit of the test also looks better here

PR #63, opened 2022-09-20
fix various bugs

kne42:
- fix typo: `KeyCode.UNKOWN` -> `KeyCode.UNKNOWN`
- add hashing to keybinding classes so they can be used as `dict` keys
- change string representations of arrow keys to better align with what's used as the standard, e.g. `UpArrow` -> `Up`


tlambert03 (approved):

tlambert03:
Since I worked hard to keep this lib at really high test coverage (currently 99%), tests are set to fail if the PR reduces test coverage.  Can you add tests for the new hash methods?

PR #64, opened 2022-09-20
fix: properly detect when ctrl and meta swapped on mac

kne42:
test the `QCoreApplication` to check if `Ctrl` and `Meta` have been swapped according to the settings


src/app_model/backends/qt/_qkeymap.py
```diff
@@ -76,109 +89,111 @@ def __init__(self, kb: KeyBinding) -> None:
         super().__init__(*ints)
 
 
+qt_keys = Qt.Key if PYQT6 else Qt
```
tlambert03, line 92: can you comment on the need for this?  tests do cover this file, and run on PySide 2/6 PyQt5/6 from 5.12 on (which is when the scoped Qt.Key was introduced) ... where were you running into issues

kne42: ah, i was copying this [from vispy](https://github.com/vispy/vispy/blob/6b9d39c8c7ab75ceb7f15ccccef5f0d27eaeeec1/vispy/app/backends/_qt.py#L153), i thought that they were right since the keys didn't appear in `dir(Qt.Key)` but upon further inspection they are implemented in the `__getattr__`. will revert

src/app_model/backends/qt/_qkeymap.py
```diff
@@ -18,13 +18,26 @@
...
+        return False
+
+    app = QCoreApplication.instance()
+    if app is None:
+        return False
+
+    return not app.testAttribute(Qt.AA_MacDontSwapCtrlAndMeta)
+
+
+def _get_qmod_lookup() -> Dict[str, Qt.KeyboardModifier]:
```
tlambert03, line 34: It would be nice if this didn't have to run on every key press, particularly on non mac systems.  Can you refactor this so that the meta_swapped function is called at most once per key press on a mac, and never on a non mac?  (Non mac shouldn't need anything besides a hash map lookup)

kne42: changed this in two places, lmk if that's what you wanted or not

src/app_model/backends/qt/_qkeymap.py
```diff
@@ -19,21 +19,45 @@
...
+def mac_ctrl_meta_swapped() -> bool:
+    """Whether or not Qt has swapped ctrl and meta for Macs."""
+    if not MAC:
+        return False
+
+    app = QCoreApplication.instance()
+    if app is None:
+        return False
+
+    return not app.testAttribute(Qt.AA_MacDontSwapCtrlAndMeta)
```
tlambert03, line 40: I believe this can just be:

```suggestion
    return not QCoreApplication.testAttribute(Qt.AA_MacDontSwapCtrlAndMeta)
```

since you're no longer calling it unless you've already checked `is MAC`, the first check is moot, and you can use `testAttribute` as a classmethod with or without an app instance.

tlambert03: in which case... perhaps just put the whole thing in _mac_qmod_lookup... and save yourself a python function call :)

src/app_model/backends/qt/_qkeymap.py
tlambert03: works for me.  could later possible do two function definitions and safe the conditional on non mac platforms altogether

src/app_model/backends/qt/_qkeymap.py
kne42: this is used inside `qkeysequence2modelkeybinding` as well and i dont think one function call is gonna slow things down that much

tests/test_qt/test_qkeymap.py
```diff
@@ -28,14 +30,32 @@ def test_qkeysequence2modelkeybinding() -> None:
...
-    )
-    app_key = KeyBinding(parts=[KeyMod.CtrlCmd | KeyCode.KeyM, KeyCode.KeyK])
-    assert qkeysequence2modelkeybinding(seq) == app_key
-
-    seq = QKeySequence(
-        Qt.Modifier.CTRL | Qt.Key.Key_M, Qt.KeyboardModifier.NoModifier | Qt.Key.Key_K
-    )
-    app_key = KeyBinding(parts=[KeyMod.WinCtrl | KeyCode.KeyM, KeyCode.KeyK])
-    assert qkeysequence2modelkeybinding(seq) == app_key
+    with patch.object(_qkeymap, "mac_ctrl_meta_swapped", return_value=True):
```
tlambert03, line 33: could just directly use `QCoreApplication.setAttribute(Qt.AA_MacDontSwapCtrlAndMeta, False)` to make it clearer what this patch is accomplishing?  (then undo at the end of the test... )

src/app_model/backends/qt/_qkeymap.py
kne42: i considered that, if we want to optimize this much tho we might as well be using cython/numba

tests/test_qt/test_qkeymap.py
kne42: ill think about it, might manually set it to a true and then a false as part of the test

src/app_model/backends/qt/_qkeymap.py
tlambert03: not sure I agree with that.  it's much easier to just avoid a python function call in the first place than to put it into cython/numba?  anyway, do what you like here.  i'm fine with it

kne42: i meant more generally if we want to apply the mentality of "this is keypress stuff that gets called a lot let's make it as efficient as possible" the best way would be cython/numba

kne42: another way i could put it is if optimization is that much more important than readability, we might as well use compilers

tests/test_qt/test_qkeymap.py
kne42: for some reason doing this messes with the other tests even tho i reset things at the end so im not going thru with it

```
________________________________ test_shortcuts ________________________________

qtbot = <pytestqt.qtbot.QtBot object at 0x7faaea3693d0>
full_app = Application('complete_test_app')

    @pytest.mark.filterwarnings("ignore:QPixmapCache.find:")
    @pytest.mark.skipif(LINUX, reason="Linux keytest not working")
    def test_shortcuts(qtbot: "QtBot", full_app: "FullApp") -> None:
        app = full_app

        win = QMainWindow()
        menu = QModelMenu(app.Menus.EDIT, app=app, title="Edit", parent=win)
        win.menuBar().addMenu(menu)
        qtbot.addWidget(win)
        qtbot.addWidget(menu)

        with qtbot.waitExposed(win):
            win.show()

        copy_action = menu.findAction(app.Commands.COPY)

        with qtbot.waitSignal(copy_action.triggered, timeout=1000):
>           qtbot.keyClicks(win, "C", Qt.KeyboardModifier.ControlModifier)
E           pytestqt.exceptions.TimeoutError: Signal triggered(bool) not emitted after 1000 ms

tests/test_qt/test_qmenu.py:115: TimeoutError
```

kne42:
@tlambert03 the following code from `SimpleKeyBinding` seems to do swapping for macs as well:

```python
@classmethod
def from_int(
    cls, key_int: int, os: Optional[OperatingSystem] = None
) -> "SimpleKeyBinding":
    """Create a SimpleKeyBinding from an integer."""
    ctrl_cmd = bool(key_int & KeyMod.CtrlCmd)
    win_ctrl = bool(key_int & KeyMod.WinCtrl)
    shift = bool(key_int & KeyMod.Shift)
    alt = bool(key_int & KeyMod.Alt)

    os = OperatingSystem.current() if os is None else os
    ctrl = win_ctrl if os.is_mac else ctrl_cmd
    meta = ctrl_cmd if os.is_mac else win_ctrl
    key = key_int & 0x000000FF  # keycode mask

    return cls(ctrl=ctrl, shift=shift, alt=alt, meta=meta, key=key)

```

since this swap is really an implementation detail of Qt do we actually want to do a swap here? also i find it weird that we perform the swap for `int` inputs but not `str` inputs

tlambert03:
> since this swap is really an implementation detail of Qt do we actually want to do a swap here?

I don't see this as an implementation detail of Qt, or related at all to Qt really...
The point is that `KeyMod.CtrlCmd` (the integer) is an os-independent concept that means "the more commonly used meta modifier for the given platform".  

https://github.com/napari/app-model/blob/7453e7b38ea0e0fdc5a8c0263508d092d684a59e/src/app_model/types/_keys/_key_codes.py#L611-L618

this lets us internally use `KeyMod.CtrlCmd` everywhere, not worrying about which platform we're on.  `SimpleKeybinding` on the other hand _is_ OS-aware.

`SimpleKeybinding` performs the appropriate swap for the current platform.  (Qt only comes into this when converting from an actual qkey-press to our internal representation...)

This is definitely complicated stuff ... I recommend looking through these two vscode files (and linked types) for inspiration on how it all fits together in a very flexible way:
https://github.com/microsoft/vscode/blob/main/src/vs/base/common/keyCodes.ts
https://github.com/microsoft/vscode/blob/main/src/vs/base/common/keybindings.ts

(`from_int` is essentially performing the same functionality as their `createSimpleKeybinding` function [here](https://github.com/microsoft/vscode/blob/main/src/vs/base/common/keybindings.ts#L46))

> also i find it weird that we perform the swap for int inputs but not str inputs

strings are only used for ingesting _external_ input (i.e. from a plugin).  where they [have the ability to specify platforms anyway](https://github.com/napari/npe2/blob/74ef51cc5933f496a43fb1538bde3b673de8711c/npe2/manifest/contributions/_keybindings.py#L19-L22).  so this is pretty much just like the VScode docs [here](https://code.visualstudio.com/docs/getstarted/keybindings#_accepted-keys)... where [external contributions are encouraged to be explicit about keys when the differ between platforms](https://code.visualstudio.com/api/references/contribution-points#keybinding-example):
```json
{
  "contributes": {
    "keybindings": [
      {
        "command": "extension.sayHello",
        "key": "ctrl+f1",
        "mac": "cmd+f1",
        "when": "editorTextFocus"
      }
    ]
  }
}
```

kne42:
ah that's very helpful, thank you. i was wondering where you got all of this stuff from so it's good to know the source/inspiration

interestingly enough as in #63 i changed `UpArrow` to `Up` for example, i was wondering where you got `UpArrow` from and it seems to be from the file you linked, but there was another field `usUserSettingsLabel` in that big matrix of keycode definitions that seems to correlate with what's output to `keybindings.json`

from a brief glance of that table, only the arrow keys seemingly have a non-`empty` `usUserSettingsLabel` that differs from the `keyCodeStr`, stripping away e.g. `UpArrow` to `Up`. @tlambert03 do you think it's worth adding this extra label and reverting the keycode string to be in line with what vscode uses, or do you think it's fine?

tlambert03:
>  i was wondering where you got all of this stuff from so it's good to know the source/inspiration

oh!  yes, I thought it was clear at this point that all of this is directly inspired/modeled after the vscode application model.

For pretty much _everything_ here, you'll find a correlate in the vscode source code.  Whenever I've had questions about how something could/should work, inasmuch as there is a correlate in vscode, I've gone to their source code to look at the patterns and translate the bits that I think we need.  Even the original [proposal of a static manifest](https://github.com/napari/napari/issues/2975) two years ago was a result of looking at the vscode patterns, and if you look through their extension-handling source code you'll find a ton of correlates with npe2.   app-model then extends that pattern to the core application itself (which is why the schema of app-model naturally/closely follows the schema of npe2).

The registries have correlates too.  so you should definitely look at the vscode keybinding registry code when it comes time to think about default < extension < user keybinding priorities.

I know there's talk of working on a command palette, so you or others should also definitely look at vscode's command palette code to look for patterns (app-model itself was really just a detour on the path to even be able to create an efficient command palette... so that was always one of the central goals here).

> @tlambert03 do you think it's worth adding this extra label and reverting the keycode string to be in line with what vscode uses, or do you think it's fine?

I noticed that, and was just gonna let you take it where you want it.  For me, since i spend a lot of time looking through the vscode source code, direct parity has been useful.  If you want to change things that's fine.  That file you changed ultimately was the internal model (not an external model)... yes, changes there might affect the external model (such as conversion from string for example), but I do think there was value A) in the parity with vscode and B) with the fact that every other row there had the string version the same as the internal model:

```
_KM(..., KeyCode.Delete, 'Delete', 46, 'VK_DELETE'),
_KM(..., KeyCode.End, 'End', 35, 'VK_END'),
_KM(..., KeyCode.PageDown, 'PageDown', 34, '
_KM(..., KeyCode.RightArrow, 'Right', 39, 'VK_RIGHT'),  # !
```

so, I wouldn't have changed it, but I'm also not going to nit-pick/micro-managing stuff going forward here.

PR #65, opened 2022-09-29
feat: use custom root type for KeyBinding

kne42:
Use a custom root type so that `models_as_dict=False` does not need to be passed to the `.json()` method since [this parameter does not get passed recursively along models](https://napari.zulipchat.com/#narrow/stream/212875-general/topic/pydantic.20model.20serialization/near/300440462). This changes the `.dict()` method to replace `KeyBinding` with its `__root__`, which is a `str`, thus effectively having the data represented as a string for serialization purposes while still being able to access the rest of the functionality attached to the `KeyBinding` class.

If I have implemented this correctly, there should be no effective changes to the API.

### DIFF
--- a/src/app_model/types/_keys/_keybindings.py
+++ b/src/app_model/types/_keys/_keybindings.py
@@ -1,7 +1,9 @@
 import re
-from typing import Any, Callable, Dict, Generator, List, Optional, Tuple
+from typing import Any, Callable, Dict, Generator, List, Optional, Tuple, no_type_check
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, PrivateAttr
+from pydantic.error_wrappers import ErrorWrapper, ValidationError
+from pydantic.errors import ListError, ListMinLengthError
 
 from .._constants import OperatingSystem
 from ._key_codes import KeyChord, KeyCode, KeyMod
@@ -136,10 +138,70 @@ class KeyBinding(BaseModel):
     the two keypress codes with a space. For example, 'Ctrl+K Ctrl+C'.
     """
 
-    parts: List[SimpleKeyBinding] = Field(..., min_items=1)
+    __root__: str
+    _parts: List[SimpleKeyBinding] = PrivateAttr()
+
+    def __init__(self, **data: Any) -> None:
+        if "parts" in data:
+            parts = data.pop("parts")
+            if not isinstance(parts, List):
+                raise ValidationError(
+                    errors=[ErrorWrapper(exc=ListError(), loc=("parts",))],
+                    model=self.__class__,
+                )
+            parts = [SimpleKeyBinding.validate(part) for part in parts]
+        elif "__root__" in data:
+            root = data["__root__"]
+            parts = [SimpleKeyBinding.from_str(part) for part in root.split()]
+        else:
+            parts = []
+
+        if len(parts) < 1:
+            raise ValidationError(
+                errors=[
+                    ErrorWrapper(ListMinLengthError(limit_value=1), loc=("parts",))
+                ],
+                model=self.__class__,
+            )
+        data["__root__"] = " ".join(str(part) for part in parts)
+
+        super().__init__(**data)
+        self._parts = parts
+
+    @property
+    def parts(self) -> List[SimpleKeyBinding]:
+        """Key combinations that make up the overall key chord."""
+        return self._parts
+
+    @no_type_check
+    def __setattr__(self, key, val):
+        if key == "__root__":
+            parts = [SimpleKeyBinding.from_str(part) for part in val.split()]
+        elif key == "parts":
+            parts = val
+            if not isinstance(parts, List):
+                raise ValidationError(
+                    errors=[ErrorWrapper(exc=ListError(), loc=("parts",))],
+                    model=self.__class__,
+                )
+            parts = [SimpleKeyBinding.validate(part) for part in parts]
+        else:
+            return super().__setattr__(key, val)
+
+        # key will always be either '__root__' or 'parts'
+        if len(parts) < 1:
+            raise ValidationError(
+                errors=[
+                    ErrorWrapper(ListMinLengthError(limit_value=1), loc=("parts",))
+                ],
+                model=self.__class__,
+            )
+
+        self._parts = parts
+        return super().__setattr__("__root__", " ".join(str(part) for part in parts))
 
     def __str__(self) -> str:
-        return " ".join(str(part) for part in self.parts)
+        return self.__root__
 
     def __eq__(self, other: Any) -> bool:
         if not isinstance(other, KeyBinding):
@@ -163,8 +225,7 @@ class KeyBinding(BaseModel):
     @classmethod
     def from_str(cls, key_str: str) -> "KeyBinding":
         """Parse a string into a SimpleKeyBinding."""
-        parts = [SimpleKeyBinding.from_str(part) for part in key_str.split()]
-        return cls(parts=parts)
+        return cls(__root__=key_str)
 
     @classmethod
     def from_int(
--- a/src/app_model/types/_keys/_keybindings.py
+++ b/src/app_model/types/_keys/_keybindings.py
@@ -3,7 +3,7 @@ from typing import Any, Callable, Dict, Generator, List, Optional, Tuple, no_typ
 
 from pydantic import BaseModel, PrivateAttr
 from pydantic.error_wrappers import ErrorWrapper, ValidationError
-from pydantic.errors import ListError, ListMinLengthError
+from pydantic.errors import ListError, ListMinLengthError, MissingError
 
 from .._constants import OperatingSystem
 from ._key_codes import KeyChord, KeyCode, KeyMod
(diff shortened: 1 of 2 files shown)


src/app_model/types/_keys/_keybindings.py
```diff
@@ -136,10 +138,60 @@ class KeyBinding(BaseModel):
...
+        super().__init__(**data)
+        self._parts = parts
+
+    @property
+    def parts(self) -> List[SimpleKeyBinding]:
+        """Key combinations that make up the overall key chord."""
+        return self._parts
+
+    @no_type_check
+    def __setattr__(self, key, val):
```
andy-sweet, line 175: Idea: do we need to support mutation of `KeyBinding`? This adds extra code to maintain, which is mostly just a repeat of `__init__` anyway.

I don't think performance is a serious concern here right? Especially since `parts` is likely to remain the only member of this thing anyway.

If we made this model frozen/immutable, then we could just derive the `parts` property (likely as a cached property) and maybe add a factory method like `KeyBinding.from_parts` to avoid some of the complexity in `__init__` too.

tests/test_keybindings.py
```diff
@@ -100,6 +117,37 @@ def test_in_dict():
     assert kbs[hash(new_a)] == 0
 
 
+def test_errors():
```
andy-sweet, line 120: Love the test coverage of error cases.

src/app_model/types/_keys/_keybindings.py
```diff
@@ -136,10 +138,60 @@ class KeyBinding(BaseModel):
...
+    __root__: str
+    _parts: List[SimpleKeyBinding] = PrivateAttr()
+
+    @classmethod
+    def _parts_error(cls, exc: Exception) -> ValidationError:
+        return ValidationError(
+            errors=[ErrorWrapper(exc=exc, loc=("parts",))], model=cls
+        )
+
+    def __init__(self, **data: Any) -> None:
```
andy-sweet, line 150: Optional: any strong reason to use `**data` instead of expanding the two known parameters `__root__` and `parts` as named keyword arguments with typing info?

andy-sweet (approved):
I can't say I love the extra code + complexity here, but I also don't have any better ideas, so approving because the desired behavior is at least well tested.

I had one optional idea around mutability that might simplify things, but not sure if it's viable.

tlambert03:
> I can't say I love the extra code + complexity here, but I also don't have any better ideas, so approving because the desired behavior is at least well tested.

I agree, while I also think the goal is reasonable (not having to pass `models_as_dict` to the `json()` method), I think what the implementation here does to the code is unfortunate. 

> I had one optional idea around mutability that might simplify things, but not sure if it's viable.

I think making this immutable is a great idea.  In any case, even if it's not immutable, I dislike the duplicated logic in the `__init__` and in `setattr`.  (That's kind what `@validator` is there for in the first place... so call it in the `__init__` if you need, rather duplicating its logic elswhere)

> Also, to confirm, the reason we want KeyBinding serialized to a string (rather than a list of dicts) is because we want it to human-readable and modifiable in something like a JSON or YAML file?

This is also my main question here: while I can see that it's a nicer way to serialize it, can you point out where you ran into this problem?  i.e. where you were trying to serialize it (rather than, say, using `str(keybinding)`) such that you found all this necessary?

src/app_model/types/_keys/_keybindings.py
```diff
@@ -136,10 +138,60 @@ class KeyBinding(BaseModel):
...
+
+        super().__init__(**data)
+        self._parts = parts
+
+    @property
+    def parts(self) -> List[SimpleKeyBinding]:
+        """Key combinations that make up the overall key chord."""
+        return self._parts
+
+    @no_type_check
```
tlambert03, line 174: please type stuff in this library

tlambert03:
if you do want to change the root type of `KeyBinding` to `str` (which I'm not convinced is necessary/advisable) then I think that `parts` should probably just be made a `property`:

```python

    __root__: str

    @property
    def parts(self) -> List[SimpleKeyBinding]:
        # could cache this and make keybinding immutable
        return [SimpleKeyBinding.from_str(part) for part in self.__root__.split()]

```

kne42:
> This is also my main question here: while I can see that it's a nicer way to serialize it, can you point out where you ran into this problem? i.e. where you were trying to serialize it (rather than, say, using str(keybinding)) such that you found all this necessary?
> 
> (an alternative very simple solution here is to set the json encoder on the model that contains the keybinding, so it would be nice to know why/where that approach is insufficient)

You can find the issue I ran into in [this Zulip thread](https://napari.zulipchat.com/#narrow/stream/212875-general/topic/pydantic.20model.20serialization/near/300440462)

Here's a simple example you can run yourself if you're not convinced:
```python
In [1]: from pydantic import BaseModel

In [2]: class Client(BaseModel):
   ...:     title: str
   ...:     first: str
   ...:     last: str
   ...:

In [3]: class Lawyer(BaseModel):
   ...:     client: Client
   ...:
   ...:     class Config:
   ...:         json_encoders = {
   ...:             Client: lambda c: f'{c.title} {c.first} {c.last}'
   ...:         }
   ...:

In [4]: class Office(BaseModel):
   ...:     lawyer: Lawyer
   ...:
   ...:     class Config:
   ...:         json_encoders = {
   ...:             Client: lambda c: f'{c.title} {c.first} {c.last}'
   ...:         }
   ...:

In [5]: client = Client(title='Dr.', first='Medikal', last='School')

In [6]: client.json(models_as_dict=False)
Out[6]: '{"title": "Dr.", "first": "Medikal", "last": "School"}'

In [7]: lawyer = Lawyer(client=client)

In [8]: lawyer.json(models_as_dict=False)
Out[8]: '{"client": "Dr. Medikal School"}'

In [9]: office = Office(lawyer=lawyer)

In [10]: office.json(models_as_dict=False)
Out[10]: '{"lawyer": {"client": {"title": "Dr.", "first": "Medikal", "last": "School"}}}'
```

tlambert03:
I don't need convincing of pydantic's behavior :) I'm aware of it.

I guess, I'm kinda asking where exactly in _your_ code you needed this. I did read the zulip thread and don't see an application there, just more MREs... and a link to the github issue where you said:

> There's an issue serializing KeyBinding to json, since to properly use the json_encoders config settings on other models, you must pass models_as_dict=False (see https://github.com/pydantic/pydantic/pull/3542). However, this aspect doesn't seem to be passed between json calls, so neither overriding the method to change the default on ShortcutSettings nor calling get_settings().json(models_as_dict=False) fixes the problem.

but what I'm looking for is the literal place where you're *using* keybinding.

kne42:
Ohhhhh, I see what you’re getting at. The equivalent of `Office` in our use case would be [`NapariSettings`](https://github.com/napari/napari/blob/main/napari/settings/_napari_settings.py). I didn’t consider adding something to encode `ShortcutSettings` as a way to properly encode `KeyBinding`.

When I have time I’ll try adding a `_json_encode` method to `ShortcutSettings` and see if that fixes things :)

tlambert03:
yeah, give it a shot.  I do still agree that it's frustrating that you can't have the logic that declares how a field should be json encoded _on_ the field itself... rather than needing to take special action in the higher level models that use the field.

Note that https://github.com/napari/napari/pull/2297 was trying to solve this internally for napari (e.g. if you define a `_json_encode`) method on the field type itself, napari's EventedModel will find it and use it.  ... not 100% sure if it will work deeply nested here, but this is definitely a case where that was "supposed" to help.  So you might also try that

kne42:
I made a [branch](https://github.com/kne42/napari/tree/test-kb) experimenting with the above, but ran into an issue with representing fully-default models. Tried to debug but just more and more issues keep popping up due to our custom handling of YAML among other things. I think at this point it would be simpler to proceed with this PR with some of your requested changes, so that's what I'll do.

potating-potato:
If I understand correctly this would be superseded by #68?

kne42:
yes

tlambert03:
@kne42  can I close this now with #68 ?

tlambert03:
closing... but feel free to reopen if I'm mistaken!

kne42:
Re-opening this, ran into some more issues where we couldn't generate the schema from `KeyBinding` since it "cannot be serialized as JSON" again as the schema function does not actually take into account the encoders we provide (perhaps another pydantic bug?). That was my bad for not running additional tests.

Do you think another solution would be better? We could theoretically subclass `KeyBinding` from `str` but things would get quite messy/confusing. Also, the latest implementation with the custom root type errors when it cannot parse a valid KeyCombo whereas previous versions would still store the parts as `Key.UNKNOWN`, is this desirable functionality?

tlambert03:
just curious, how did using napari's `_json_encode` work out?

tlambert03:
more context about the issues you're running into would also be appreciated.  minimal examples using pydantic are helpful, but i know that there are lots of little tricky bits with pydantic, so seeing the actual code and problems you're running into would be very helpful here so we're not just talking in the abstract

tlambert03:
> when it comes to generating the schema, they... just use their default one

ohhh... I see, your issue is in `schema_json()`... did you try `__modify_schema__`?
https://pydantic-docs.helpmanual.io/usage/schema/#modifying-schema-in-custom-fields

> im at the point where i just want to rewrite a ton of pydantics source code

well you're in luck, they're hard at work on v2, re-written in rust.

tlambert03:
as a side comment here: I also think that using the json-schema autogenerator thing in the napari preferences dialog has been stretched well past its original scope (and usefulness) ... so another thing you could consider here is simply excluding this from the schema there, and creating a custom widget without going through json-schema-form-thingy

kne42:
> ohhh... I see, your issue is in `schema_json()`... did you try `__modify_schema__`? https://pydantic-docs.helpmanual.io/usage/schema/#modifying-schema-in-custom-fields

yes, turns out that's called later down the chain after it's already been passed to their default encoding function which [throws an error](https://github.com/pydantic/pydantic/blob/02d4e188395cb6c408625bf001ee18dbd9ecc1fb/pydantic/json.py#L90)

> well you're in luck, they're hard at work on v2, re-written in rust.

thank god im actually at my wit's end with this library's design, perhaps it's time to learn rust and contribute 😓

tlambert03:
yes, no doubt there are frustrations.  but i think on the whole, it's an excellent idea and library, with some hiccups in the original implementation.  (something I think we should all be able to relate to as napari developers :joy:) each time I find myself thinking I could do better, I find it harder than at first I thought

PR #70, opened 2022-10-12
feat: use custom root type for KeyBinding

kne42:
it won't let me re-open the previous PR since I force-pushed so see #65 for context

### DIFF
--- a/src/app_model/types/_keys/_keybindings.py
+++ b/src/app_model/types/_keys/_keybindings.py
@@ -1,7 +1,9 @@
 import re
-from typing import Any, Callable, Dict, Generator, List, Optional, Tuple
+from typing import Any, Callable, Dict, Generator, List, Optional, Tuple, no_type_check
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, PrivateAttr
+from pydantic.error_wrappers import ErrorWrapper, ValidationError
+from pydantic.errors import ListError, ListMinLengthError
 
 from .._constants import OperatingSystem
 from ._key_codes import KeyChord, KeyCode, KeyMod
@@ -125,7 +127,7 @@ class SimpleKeyBinding(BaseModel):
         return super().validate(v)
 
 
-class KeyBinding:
+class KeyBinding(BaseModel):
     """KeyBinding.  May be a multi-part "Chord" (e.g. 'Ctrl+K Ctrl+C').
 
     This is the primary representation of a fully resolved keybinding. For consistency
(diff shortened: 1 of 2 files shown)


tlambert03:
ok, I took a look deeper at the failing test, and I have to say, I still just feel like we're focusing on the app-model side more than the napari side here.  The issue (as I'm sure you're well aware, but now I am too) is that pydantic is now having difficulty serializing the _default_ value for `ShortcutsSettings.shortcuts` ... (not the field itself) which, after #68 was merged, is now essentially a dict of arbitrary objects.  which indeed does pose more of a challenge for serialization than it did when it was a base model.

looking through the field default serialization code in pydantic, it looks like the main reason that `__modify_schema__` didn't appear to be working for you is that it's actually not making it to that point before it chokes on serializing the default value.  However, one way to avoid trying to serialize a complex default value when generating the schema is simply to not provide it to `default`, but to use a `default_factory instead`.  (then you'll bypass the [`encode_default` function here](https://github.com/pydantic/pydantic/blob/02d4e188395cb6c408625bf001ee18dbd9ecc1fb/pydantic/schema.py#L214-L216))

So, before we continue doing surgery on the Keybindings field (which really is and should be able to be a rather simple field)... I'd look harder at `ShortcutSettings`.  Try, for example, to use:

```
class ShortcutsSettings(EventedModel):
    shortcuts: Dict[str, List[KeyBinding]] = Field(
        default_factory=lambda: default_shortcuts,
        ...
```

and _then_ either revert #68 or use `modify_schema` with something like

```
    @classmethod
    def __modify_schema__(cls, field_schema, field: Optional['ModelField']):
        if field:
            field_schema['type'] = 'string'
```

kne42:
okay, thanks for investigating and coming up with alternative solutions, ill try this one out today and let you know how it goes

kne42:
so it does fix the schema bug, however it then doesnt properly factor into the logic for excluding defaults since it is a factory, will investigate further

```python
 assert safe_load(fake_path.read_text()) == {
            'schema_version': CURRENT_SCHEMA_VERSION
        }
E       AssertionError: assert {'schema_vers... ['P'], ...}}} == {'schema_vers..., build=None)}
E         Omitting 1 identical items, use -vv to show
E         Left contains 1 more item:
E         {'shortcuts': {'shortcuts': {'napari:activate_add_ellipse_mode': ['E'], ...
```

kne42:
Interesting. If your main issue is with putting the hacky workarounds inside `app-model`, how would you feel about doing the same hacky stuff you hate but putting it in napari instead? I could make a wrapper around the `app_model.KeyBinding` object and do all the surgery on that and stick it in the napari repo itself.

tlambert03:
> how would you feel about doing the same hacky stuff you hate but putting it in napari instead?

i would like that more!  I do still think these workarounds are indicative of deeper problems on the napari side... but given that those problems already exist there, and less so here, i'd much rather keep them "grouped" together over there :joy:

PR #73, opened 2022-10-17
represent keybind as string

kne42:
alternative to #65


potating-potato:
I am also in favor of extending this as a list than str given its nature. There was some minor issue on the string representation but we can always override tostring method to deal with that

potating-potato:
with the workaround on napari side I think this can be closed?

PR #76, opened 2022-11-08
fix: fix unsupported operand

tlambert03:
fixes `unsupported operand type(s) for &: 'PySide6.QtCore.QKeyCombination' and 'KeyboardModifier'` on the newest pyside6

side-note: fixes in https://github.com/napari/superqt/pull/133 should get pyside6 tests working again


aganders3 (approved):
This looks good to me. Tests pass locally (macOS) with both PyQt6 and PySide6.

One nit is that `PYQT6` is now unused in this file so can be removed from the import.

PR #82, opened 2022-11-18
fix: Fix ctrl meta key swap (for real this time (i think))

kne42:
yeah so i made an oopsie in https://github.com/pyapp-kit/app-model/pull/64 and only had the logic checking for modifier swaps... it now properly handles keycode swaps as well 😅

### DIFF
--- a/src/app_model/backends/qt/_qkeymap.py
+++ b/src/app_model/backends/qt/_qkeymap.py
@@ -256,7 +256,7 @@ def qmods2modelmods(modifiers: Qt.KeyboardModifier) -> KeyMod:
     lookup = SWAPPED_KEYMOD_FROM_QT if mac_ctrl_meta_swapped() else KEYMOD_FROM_QT
     for modifier in lookup:
         if modifiers & modifier:
-            mod |= modifier
+            mod |= lookup[modifier]
     return mod
 
 
--- a/src/app_model/backends/qt/_qkeymap.py
+++ b/src/app_model/backends/qt/_qkeymap.py
@@ -40,7 +40,7 @@ _SWAPPED_QMOD_LOOKUP = {
 }
 
 
-def qt_swapped_ctrl_meta() -> bool:
+def mac_ctrl_meta_swapped() -> bool:
     """Determine whether Qt is swapping Ctrl and Meta for keyboard interactions."""
     return not QCoreApplication.testAttribute(Qt.AA_MacDontSwapCtrlAndMeta)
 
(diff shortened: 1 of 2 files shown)


tlambert03 (approved):
looks good @kne42  thanks.

> it now properly handles keycode swaps as well

Is there a test you could add that would capture what was actually broken here?

tlambert03:
hi @kne42 ... this was approved back in December, but didn't get a response regarding adding a test that would have caught this.  do you still want this in?  If so, please let me know about that last question

src/app_model/backends/qt/_qkeymap.py
```diff
@@ -250,30 +264,33 @@ def __init__(self, kb: KeyBinding) -> None:
...
-    if modifiers & Qt.KeyboardModifier.MetaModifier:
-        if swapped is None:
-            swapped = MAC and mac_ctrl_meta_swapped()
-        if swapped:
-            mod |= KeyMod.CtrlCmd
-        else:
-            mod |= KeyMod.WinCtrl
+    lookup = (
+        MAC_KEYMOD_FROM_QT if MAC and not mac_ctrl_meta_swapped() else KEYMOD_FROM_QT
+    )
```
Czaki, line 270: I do not like this if. It could be resolved on import time. Execution of this if every function call looks like a bad pattern.

src/app_model/backends/qt/_qkeymap.py
```diff
@@ -19,41 +19,41 @@
...
+    **_QMOD_LOOKUP,
+    "ctrl": QMETA,
+    "meta": QCTRL,
+}
+
 
 def mac_ctrl_meta_swapped() -> bool:
-    """Whether or not Qt has swapped ctrl and meta for Macs."""
+    """Determine whether Qt is swapping Ctrl and Meta for keyboard interactions."""
     return not QCoreApplication.testAttribute(Qt.AA_MacDontSwapCtrlAndMeta)
```
Czaki, line 45: Is this could be changed during execution time, or it could be resolved once?

Czaki (commented):
@tlambert03 Could you get another look as it is required for https://github.com/napari/napari/pull/5604?

src/app_model/backends/qt/_qkeymap.py
tlambert03: I'm not sure that's true?  `Qt.AA_MacDontSwapCtrlAndMeta` can change dynamically.

```python
In [3]: from qtpy.QtCore import Qt, QCoreApplication

In [6]: app = QCoreApplication([])

In [7]: app.testAttribute(Qt.AA_MacDontSwapCtrlAndMeta)
Out[7]: False

In [8]: app.setAttribute(Qt.AA_MacDontSwapCtrlAndMeta)

In [9]: app.testAttribute(Qt.AA_MacDontSwapCtrlAndMeta)
Out[9]: True
```

i'm ok leaving that possible performance improvement for later

PR #94, opened 2023-04-04
refactor: Fix type annotation in action

Czaki:
Based on usage in napari:

https://github.com/napari/napari/blob/6416098946d594989072c2044551d7e0eca08328/napari/_app_model/actions/_layer_actions.py#L47

To avoid such messages 

obraz

### DIFF
--- a/src/app_model/types/_action.py
+++ b/src/app_model/types/_action.py
@@ -6,7 +6,7 @@ from pydantic import Field, validator
 
 from ._command_rule import CommandRule
 from ._keybinding_rule import KeyBindingRule
-from ._menu_rule import MenuRule
+from ._menu_rule import MenuRuleOrDict
 from ._utils import _validate_python_name
 
 # maintain runtime compatibility with older typing_extensions
@@ -41,7 +41,7 @@ class Action(CommandRule, Generic[P, R]):
         "`{obj.__module__}:{obj.__qualname__}` "
         "(e.g. `my_package.a_module:some_function`)",
     )
-    menus: Optional[List[MenuRule]] = Field(
+    menus: Optional[List[MenuRuleOrDict]] = Field(
         None,
         description="(Optional) Menus to which this action should be added.",
     )


Czaki:
Ok. Looking at the tests it looks like the wrong approach. Did you have any idea other than direct call `MenuRole` in the file to fix annotations?

tlambert03:
yeah, @Czaki, this is a general problem with typing in a parsing/coercing library like pydantic.  There is no way easy way to express all of the valid types that the `__init__` method *accepts* ... and we definitely don't want to change the type of the field itself (the type that it becomes coerced to). 

To say that `Action` has `menus: Optional[List[MenuRuleOrDict]]` would actually be incorrect.  because any dict inputs are parsed and converted by pydantic to `MenuRule` objects (that's kinda the whole point of pydantic in one regard).

The fact that you are allowed to pass in dicts as conveniences to generate objects is nice, but as you see, does create problems for typing.  For example:

Screen Shot 2023-04-05 at 10 52 56 AM


The solution is not to change the model itself (we *don't* want the type of the field to be a Union between dict and `MenuRule`  ... we want it to be `MenuRule`)... but if you'd like to remove the hint errors on napari side, you can either use `# type: ignore` ... or go ahead and construct the actual `app-model` `MenuRule` objects from the dicts before passing them into `Action`.

tlambert03:
i'm inclined to close this, but will let you do so when ready

tlambert03:
> Did you have any idea other than direct call MenuRole in the file to fix annotations?

this is also the kind of thing that the `pydantic.mypy` plugin could conceivably do for you, but that seems unlikely as well, and of course only works for mypy (not pyright or anything else)

tlambert03:
> Did provide overload init method could help to solve this problem?

yes, I suppose... but wouldn't you then have to do that on _every_ model?  currently, one doesn't typically make an `__init__` method in a pydantic model at all.  So that's a _lot_ of extra boilerplate to add all over the place, and a lot of stuff to keep in sync as you update the model.  (`Action` may be the place you noticed this, but this issue pops up everywhere).

Czaki:
For how Manu of models there is provided typed dictionary?

tlambert03:
> For how Manu of models there is provided typed dictionary?

This doesn't just pertain to typed dicts: *any* model's init args can be represented by a typed dictionary.  The fact that I did it for menus and keybindings there was just so that I could provide better typing for the  `register_action` function.  I'm 100% sure that I don't want to override `__init__` on pydantic methods just to indicate that you can also pass a dict.   This is a very common/old problem, not specific to app model, and the way to solve it is to either just pass the proper type to the `__init__` method of a Pydantic object, or use `type ignore` if you'd like to pass in another type and let pydantic do the type conversion.

PR #95, opened 2023-04-04
fix: don't use mixin for menus

### LINKED ISSUE #93: Error instantiating QModelMenu with PySide6 6.5.0
* app-model version: `main` (`0.1.2`)
* Python version: `3.10.10`
* Operating System: macOS 13.3

### Description
There is a problem instantiating the `QModelMenu` with PySide 6.5.0. It looks like multiple inheritance gets a bit trickier here as now QWidget is effectively calling `super().__init__`? This feels like a pretty big change so perhaps it should be reported on the Qt bug tracker.

Try this demo script to see what I mean:

```python
# qwidget_demo.py
from PySide6.QtCore import QObject
from PySide6.QtWidgets import QApplication, QWidget

app = QApplication([])


class A(QObject):
    def __init__(self, *args, **kwargs):
        print("A.__init__", args, kwargs)


class B(QWidget, A):
    def __init__(self, *args, **kwargs):
        QWidget.__init__(self, *args, **kwargs)
        print("B.__init__", args, kwargs)


print("instantiating A")
A()

print("instantiating B")
B()
```

#### output with PySide 6.4.2
```
❯ python qwidget_demo.py 
instantiating A
A.__init__ () {}
instantiating B
B.__init__ () {}
```
#### output with PySide 6.5.0
```
❯ python qwidget_demo.py 
instantiating A
A.__init__ () {}
instantiating B
A.__init__ () {}
B.__init__ () {}
```

### What I Did
Create a virtual environment with the newly released PySide6 v6.5.0:
```
% python -m venv venv
% source venv/bin/activate
% pip install pyside6==6.5.0
% pip install -e '.[qt,test,test-qt]'
% pytest
```

Tests are failing for `QMenu` with this rather confusing output:

```
self = <[RuntimeError("'__init__' method of object's base class (QModelMenu) not called.") raised in repr()] QModelMenu object at 0x12e46ae00>, menu_id = 'help'
app = Application('complete_test_app'), title = None, parent = None

    def __init__(
        self,
        menu_id: str,
        app: Union[str, Application],
        title: Optional[str] = None,
        parent: Optional[QWidget] = None,
    ):
>       QMenu.__init__(self, parent)
E       TypeError: _MenuMixin.__init__() missing 1 required positional argument: 'app'

src/app_model/backends/qt/_qmenu.py:144: TypeError
```

### Possible solution?

I'm not sure what the best solution is here - I don't have a ton of experience with multiple inheritance or mixins. One thing that seems to work is to give `_MenuMixin` a dummy `__init__` like this, but it doesn't necessarily feel like the right way to do it:

```diff
--- a/src/app_model/backends/qt/_qmenu.py
+++ b/src/app_model/backends/qt/_qmenu.py
@@ -44,7 +44,10 @@ class _MenuMixin(QObject):
     _app: Application
     _menu_id: str
 
-    def __init__(
+    def __init__(self, *args, **kwargs):
+        pass
+
+    def _init_mixin(
         self,
         menu_id: str,
         app: Union[str, Application],
@@ -141,8 +144,9 @@ class QModelMenu(QMenu, _MenuMixin):
         title: Optional[str] = None,
         parent: Optional[QWidget] = None,
     ):
+        print(QModelMenu.__mro__)
         QMenu.__init__(self, parent)
-        _MenuMixin.__init__(self, menu_id, app)
+        self._init_mixin(menu_id, app)
         if title is not None:
             self.setTitle(title)
         self.aboutToShow.connect(self._on_about_to_show)
@@ -217,7 +221,7 @@ class QModelToolBar(QToolBar, _MenuMixin):
     ) -> None:
         self._exclude = exclude
         QToolBar.__init__(self, parent)
-        _MenuMixin.__init__(self, menu_id, app)
+        self._init_mixin(menu_id, app)
         if title is not None:
             self.setWindowTitle(title)
```

tlambert03:
fixes #93 

while it helped reduce a little bit of code duplication, i was never particularly happy with the _MenuMixin anyway.  So this just removes it

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,7 +81,7 @@ jobs:
           python -m pip install ${{ matrix.qt-backend }}
 
       - name: Test
-        uses: coactions/setup-xvfb@v1
+        uses: aganders3/headless-gui@v1
         with:
           run: python -m pytest -s --cov=app_model --cov-report=xml --cov-report=term-missing --color=yes
 
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -41,7 +41,7 @@ jobs:
           python -m pip install --pre ${{ matrix.qt-backend }}
 
       - name: Test
-        uses: GabrielBB/xvfb-action@v1
+        uses: aganders3/headless-gui@v1
         with:
           run: python -m pytest -s --color=yes
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,11 +85,11 @@ line-length = 88
 src = ["src", "tests"]
 target-version = "py38"
 extend-select = [
-    "E",    # style errors
-    "F",    # flakes
-    "D",    # pydocstyle
+    "E", # style errors
+    "F", # flakes
+    "D", # pydocstyle
     "I", # isort
-    "U",    # pyupgrade
+    "U", # pyupgrade
     # "N",  # pep8-naming
     # "S",  # bandit
     "C",    # flake8-comprehensions
@@ -123,7 +123,10 @@ keep-runtime-typing = true
 # https://docs.pytest.org/en/6.2.x/customize.html
 [tool.pytest.ini_options]
 minversion = "6.0"
-filterwarnings = ["error", "ignore:QPixmapCache.find:DeprecationWarning:"]
+filterwarnings = [
+    "error",
+    "ignore:Enum value:DeprecationWarning:superqt",
+]
 
 # https://mypy.readthedocs.io/en/stable/config_file.html
 [tool.mypy]
--- a/src/app_model/backends/qt/__init__.py
+++ b/src/app_model/backends/qt/__init__.py
@@ -9,7 +9,7 @@ from ._qkeymap import (
     qmods2modelmods,
 )
 from ._qmainwindow import QModelMainWindow
-from ._qmenu import QModelMenu, QModelMenuBar, QModelSubmenu
+from ._qmenu import QModelMenu, QModelMenuBar, QModelSubmenu, QModelToolBar
 from ._util import to_qicon
 
 __all__ = [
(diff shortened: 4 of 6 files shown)


tlambert03:
@Czaki, this is now working on PySide6 everywhere but linux, where we're now seeing:

```
tests/test_qt/test_demos.py qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
[117](https://github.com/pyapp-kit/app-model/actions/runs/4613142601/jobs/8154827820?pr=95#step:6:120)
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.
[118](https://github.com/pyapp-kit/app-model/actions/runs/4613142601/jobs/8154827820?pr=95#step:6:121)

[119](https://github.com/pyapp-kit/app-model/actions/runs/4613142601/jobs/8154827820?pr=95#step:6:122)
Available platform plugins are: xcb, eglfs, offscreen, vnc, wayland, minimalegl, linuxfb, vkkhrdisplay, minimal, wayland-egl.
[120](https://github.com/pyapp-kit/app-model/actions/runs/4613142601/jobs/8154827820?pr=95#step:6:123)

[121](https://github.com/pyapp-kit/app-model/actions/runs/4613142601/jobs/8154827820?pr=95#step:6:124)
Fatal Python error: Aborted
[122](https://github.com/pyapp-kit/app-model/actions/runs/4613142601/jobs/8154827820?pr=95#step:6:125)
```

do you know if PySide 6.5.0 added library dependencies that need to be added to `setup-qt-libs`?

Czaki:
> do you know if PySide 6.5.0 added library dependencies that need to be added to `setup-qt-libs`?

I think it is a general Qt problem, and it now requires `libxcb-cursor0`. (The PyQt6 last release is 6.4.2).  I have created PR to `setup-qt-libs`

PR #98, opened 2023-05-01
feat: support pydantic v2

tlambert03:
this supports pydantic v2 (without dropping support for v1) ... but requires the `main` branch (later than the current pre-release candidate of 2.0a3)


tlambert03:
This is now working for both v1 and v2 and test coverage complete.
Will likely leave here until v2.0a4 is released, since this uses the `@model_validate` decorator in `main` that isn't yet released

PR #104, opened 2023-06-04
fix: Amend preferences `StandardKeyBinding`

lucyleeow:
[Qt](https://doc.qt.io/qt-6/qkeysequence.html#standard-shortcuts) suggests this shortcut for MacOS, with no bindings for Win or Linux. [VS code](https://code.visualstudio.com/docs/getstarted/keybindings#_keyboard-shortcuts-reference) uses this shortcut for all platforms.

I have amended primary key binding for 'preferences' to be `Ctrl + ,`. 
Another option is to follow Qt and only set keybinding to be this for Mac, leaving rest as None. Happy to change.


tlambert03:
Thanks!

PR #107, opened 2023-06-13
docs: Move `_expressions.py` docstring to be included in documentation

### LINKED ISSUE #106: DOC Add `Expr` background and supported expressions
There is some great info in the `src/app_model/expressions/_expressions.py` module docstring, which is unfortunately not included in the documentation. Would you consider including this / is there a way to easily do this in the current documentation system?

Also would a list of supported comparisons / ops be useful?

lucyleeow:
Closes #106 

Have added a 'includes ..' to note less well known operators, can remove this.

### DIFF
--- a/src/app_model/expressions/_expressions.py
+++ b/src/app_model/expressions/_expressions.py
@@ -103,8 +103,9 @@ class Expr(ast.AST, Generic[T]):
     Expressions that are supported:
     - Names: 'myvar' (these must be evaluated along with some context)
     - Constants: '1'
-    - Comparisons: 'myvar > 1' (includes `in`, `not in`)
-    - Boolean Operators: 'myvar and yourvar'
+    - Comparisons: 'myvar > 1'
+    - Boolean Operators: 'myvar & yourvar' (bitwise `&` and `|` are overloaded
+      here to mean boolean `and` and `or`)
     - Binary Operators: 'myvar + 42' (includes `//`, `@`, `^`)
     - Unary Operators: 'not myvar'
 
--- a/src/app_model/expressions/_expressions.py
+++ b/src/app_model/expressions/_expressions.py
@@ -1,28 +1,4 @@
-"""Provides the  :class:`Expr` and its subclasses.
-
-`Expr` is a subclass of `ast.AST` that provides rich dunder methods that
-facilitate joining and comparing typed expressions.  It only implements a
-subset of ast Expressions (for safety of evaluation), but provides more
-than ast.literal_eval.
-
-Expressions that are supported:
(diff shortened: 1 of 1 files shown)


src/app_model/expressions/_expressions.py
```diff
@@ -119,6 +95,29 @@ def safe_eval(expr: Union[str, bool, Expr], context: Optional[Mapping] = None) -
...
 
+    This is a subclass of `ast.AST` that provides rich dunder methods that
+    facilitate joining and comparing typed expressions.  It only implements a
+    subset of ast Expressions (for safety of evaluation), but provides more
+    than `ast.literal_eval`.
+
+    Expressions that are supported:
+    - Names: 'myvar' (these must be evaluated along with some context)
+    - Constants: '1'
+    - Comparisons: 'myvar > 1' (includes `in`, `not in`)
```
tlambert03, line 106: The include isn't actually correct... we _dont_ support the `in` operator (that would be a `__contains__` method).  So someone would specifically need to use the special `in_` and `not_in` methods on these expression objects (which is a funnier syntax).  If that doesn't make sense, probably best just to remove the parenthetical addition here

src/app_model/expressions/_expressions.py
```diff
@@ -119,6 +95,29 @@ def safe_eval(expr: Union[str, bool, Expr], context: Optional[Mapping] = None) -
...
+    - Boolean Operators: 'myvar and yourvar'
```
tlambert03, line 107: We probably need an addition comment here.  I think it's actually `&` rather than `and` here... (ie we overload the bitwise and operator to mean the Boolean and)

src/app_model/expressions/_expressions.py
lucyleeow: My bad, did not read properly. I will remove it.

Just for info, I found that `in` did seem to work:
```python
from app_model.expressions import parse_expression

a = parse_expression('1 in myvar')
a.eval({'myvar': (1,2)})
```
gives `True`. When you say we don't support it, do you mean it will 'work' but shouldn't be used as it behaves weird?

src/app_model/expressions/_expressions.py
lucyleeow: Done.

I noticed that `(() and (1,2))` should work for boolean `and` but gives an error here:
```python
a = parse_expression('a & b')
a.eval({'a':(), 'b': (1,2)})
```
```
TypeError: unsupported operand type(s) for &: 'tuple' and 'tuple'
```

tlambert03 (approved):

PR #112, opened 2023-07-09
feat: Deprecate parsing unknown keys.

Carreau:
This makes it invalid to parse win/cmd on non relevant platform, but I guess we could add placeholder values for those if we wish.

This should avoid parsing typoes

### DIFF
--- a/tests/test_keybindings.py
+++ b/tests/test_keybindings.py
@@ -46,13 +46,17 @@ def test_simple_keybinding_single_mod(mod: str, key: str) -> None:
     assert int(as_full_kb) == int(kb)
 
 
-def test_simple_keybinding_multi_mod():
+@pytest.mark.skipif(sys.platform != "darwin", reason="mac only")
+def test_simple_keybinding_multi_mod_mac():
     # here we're also testing that cmd and win get cast to 'KeyMod.CtrlCmd'
 
     kb = SimpleKeyBinding.from_str("cmd+shift+A")
     assert not kb.is_modifier_key()
     assert int(kb) & KeyMod.CtrlCmd | KeyMod.Shift
 
+
+@pytest.mark.skipif(sys.platform == "darwin", reason="win not defined on mac")
+def test_simple_keybinding_multi_mod_win():
     kb = SimpleKeyBinding.from_str("win+shift+A")
     assert not kb.is_modifier_key()
     assert int(kb) & KeyMod.CtrlCmd | KeyMod.Shift
@@ -98,6 +102,11 @@ def test_in_dict():
 
     with pytest.raises(KeyError):
         kbs[new_a]
+
+
+def test_warn_from_string_unknown():
+    with pytest.warns(DeprecationWarning):
+        KeyBinding.from_str("Shift+Return")
 
 
 def test_in_model():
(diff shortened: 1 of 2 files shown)


tests/test_keybindings.py
```diff
@@ -46,13 +46,17 @@ def test_simple_keybinding_single_mod(mod: str, key: str) -> None:
     assert int(as_full_kb) == int(kb)
 
 
-def test_simple_keybinding_multi_mod():
+@pytest.mark.skipif(sys.platform != "darwin", reason="mac only")
```
tlambert03, line 49: these skips don't seem right to me... You should absolutely be able to represent the *concept* of `"win+shift+A"` on a mac:

```py
In [1]: from app_model.types import SimpleKeyBinding

In [2]: SimpleKeyBinding.from_str("win+shift+A")
Out[2]:
SimpleKeyBinding(
    ctrl=False,
    shift=True,
    alt=False,
    meta=True,
    key=<KeyCode.KeyA: 18>
)
```


why would we lose that?

tlambert03: i don't see how this is related to your Shift+Return issue.   This was always meant to be platform agnostic at this point

Carreau: but...

```
In [3]: SimpleKeyBinding.from_str("win")
Out[3]: SimpleKeyBinding(ctrl=False, shift=False, alt=False, meta=False, key=<KeyCode.UNKNOWN: 0>)
```

tlambert03: ahh... I see, it's the ending with a modifier issue again,  ok, for that we need to look not at SimpleKeyBinding, but at KeyCode I think.  lemme take a look

Carreau: Even on windows this the win keyCode seem to be UNKNOWN

tlambert03: yeah, this is the line then that needs consideration:

https://github.com/pyapp-kit/app-model/blob/08d0fbd6274ebdcf0d2634dc0bf7f8a9b4446004/src/app_model/types/_keys/_keybindings.py#L79

specifically... we need the logic that maps "win" and "cmd" to meta:

https://github.com/pyapp-kit/app-model/blob/08d0fbd6274ebdcf0d2634dc0bf7f8a9b4446004/src/app_model/types/_keys/_keybindings.py#L289-L292

to be added to the `KeyCode.from_string` logic:

https://github.com/pyapp-kit/app-model/blob/08d0fbd6274ebdcf0d2634dc0bf7f8a9b4446004/src/app_model/types/_keys/_key_codes.py#L588-L591

tlambert03: right, because if we're talking about single keys (not key combos) then there is only the `Meta` key... 

https://github.com/pyapp-kit/app-model/blob/08d0fbd6274ebdcf0d2634dc0bf7f8a9b4446004/src/app_model/types/_keys/_key_codes.py#L93

so just need to update `KeyCode.from_string()` to map `win` and `cmd` to that

tlambert03: just FYI, in general, there isn't any "statefulness" in the key representations that pays any attention to what OS you're on until the last minute, when the keybinding actually gets registered at runtime:

https://github.com/pyapp-kit/app-model/blob/08d0fbd6274ebdcf0d2634dc0bf7f8a9b4446004/src/app_model/registries/_keybindings_reg.py#L54-L69


all the rest of the keybinding stuff is OS-agnostic

PR #113, opened 2023-07-10
feat: map win and cmd to meta

### LINKED ISSUE #111: Should Kebindings.from_string fail on invalid/unkown keys ?
* app-model version: latest

### Description

I was surprised by the following:

```
In [1]: str(KeyBinding.from_str('Meta+Shift+Return'))
Out[1]: 'Shift+Meta+'
```

A realized that `Return` is not a handled key, should we have a warning/error on invalid values ?

tlambert03:
cc @carreau ... would this fix it for you?

closes #112
closes #111


Carreau:
In the context of #112, yeah I think that works.

PR #127, opened 2023-10-19
draft: https://github.com/napari/napari/issues/6350

VolkerH:
supporting material for https://github.com/napari/napari/issues/6350

No intention to leave draft or be reviewed.

### DIFF
--- a/src/app_model/_pydantic_compat.py
+++ b/src/app_model/_pydantic_compat.py
@@ -4,7 +4,7 @@ from typing import Any, Callable, Literal, TypeVar
 
 from pydantic import BaseModel, __version__
 
-PYDANTIC2 = __version__.startswith("2")
+PYDANTIC2 = False
 M = TypeVar("M", bound=BaseModel)
 C = TypeVar("C", bound=Callable[..., Any])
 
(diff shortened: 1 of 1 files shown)


tlambert03:
I read https://github.com/napari/napari/issues/6350 ... but I still don't quite get this one.  It looks like you're prepping to be able to pin `pydantic >2` by removing the version check?  but i don't think that would be the goal of a napari update?

VolkerH:
@jni asked me to put in what I had tried in terms of making napari work
with pydantic 2.4 as draft PRs. This was just part of a quick hack to test
napari with `pydantic.v1` imports everywhere. I hard-coded pyappkit to use
pydantic v2.4 just for this single (failed) test on my local machine. This
is in no way intended to be a feature suggestion.

tlambert03:
> This is in no way intended to be a feature suggestion.

yep, I gathered that from the original message :) 
I guess I was more curious to hear why you had to do this at all to make app-model work with pydantic v2.4 ... since I would have expected app-model to already be all set (since #98)

VolkerH:
Ah... I was just trying to make sure that app-model was also using
pydantic.v1.* rather than pydantic for consistency with the other packages
I patched. I had never looked at app_model before and wasn't sure whether
the compatibility layer would work otherwise. (if it is just local use
within it shouldn't matter of course, but I wasn't sure whether napari
somehow sends/receives pydantic classes from/to app_model machinery.

tlambert03:
ah, gotcha.  thanks!

jni:
Yeah, sorry for the noise, a lot of activity happening with limited communication. I think what we want to do with napari is to allow it to work with *either* v1 or v2. Then @VolkerH you'd be unstuck with your own work because you could install v2 for your library *and* use napari, within the same env. Correct?

PR #129, opened 2023-10-27
test: run napari tests using napari workflow

tlambert03:
https://github.com/napari/napari/pull/6336


tlambert03:
@Czaki ... I'm a bit confused.

In the npe2 pr, you used:

```yml
  test_napari:
     name: napari tests
     uses: czaki/napari/.github/workflows/test_external_package_workflow.yml@feature/test_against_napari
```

- it kinda seems like `test_external_package_workflow` is just a workflow to test the real workflow which is `reusable_run_tox_test` ... 
- when I use it here, it looks like it's just pulling in npe2 (and not app-model)

is that correct?  should I be switching it to `reusable_run_tox_test` here?  and should you be switching it to that in https://github.com/napari/npe2/pull/322 as well?

tlambert03:
hmmm... but then if I do that, it doesn't test against both HEAD and release of napari, which I gathered was the point.  I'll wait for you to tell me what I've misunderstood :)

Czaki:
Short summary. 
There was a bug in workflow file (wrong if) that lead to the pulling of npe2 instead of the proper repository. It is fixed. To get a proper working workflow, it is required to restart workflows with a new commit (otherwise, GitHub will select old workflow version). I suggest `git commit -m "trigger ci" --allow-empty` 

The idea is to test against the latest release and the current main. But this workflow requires updating tox configuration to allow the install dependency from the repository. So the current version tests against branch `v0.4.19x` (current rc) and `main_test`, the branch with modification of `tox.ini`. 

When PR is merged to `napari/napari` repository it will be tested against `main` branch

PR #130, opened 2023-10-27
feat: support iconify icon keys

tlambert03:
This PR 
- takes advantage of superqt iconify support (https://github.com/pyapp-kit/superqt/pull/209) and allows icons to be expressed as iconify keys: https://icon-sets.iconify.design
  ```py
  Action(id=..., title=..., icon='fa6-solid:arrow-up')
  ```

see #132 for follow up with color stuff


tlambert03:
cc @jni, @Czaki ... nothing here affects napari, since napari still doesn't use app-model for it's icons.  But plugins could conceivable in the future if anyone starts using the [`commands.icon` spec](https://napari.org/stable/plugins/contributions.html#contributions-commands) ... this makes it easier for them to pick icons from iconify

tlambert03:
I pulled off the color spec into #132  ... not entirely sure  about that part yet.  so this is just iconify now

PR #137, opened 2023-10-31
refactor: remove comparison between Keybinding and string

### LINKED ISSUE #108: Keybinding comparing itself to `str` is really weird.
* app-model version: 0.1.4
* Python version: 3.10
* Operating System: Mac

### Description

I see what you are doing with Keybinding, and trying to make it magically work with strings, but I think it's a footgun in the long run. 

### What I Did

```
In [1]: from app_model.types import KeyBinding

In [2]: X = [KeyBinding.from_str('A')]

In [3]: [isinstance(s, str) for s in X]
Out[3]: [False]

In [4]: 'A' in X
Out[4]: True
```

For me `Out[3]` logically implies that `In [4]` Cannot be True. And equal but hash is different is a footgun as well:

```
In [7]: d = {}

In [8]: d[X[0]] = 1

In [9]: d['A'] # must be 1 as X[0] == 'A' ?
```

No `KeyError`.

I think this is make worse by the fact that `__str__` return the keybinding in the string form as now code in napari liberally use the object directly in places where it's implicitely `str()`'d.

tlambert03:
closes #108
@carreau, if you could review/confirm that this is what you wanted in #108 I'd appreciate it!

(looks like napari tests [pass fine](https://github.com/pyapp-kit/app-model/actions/runs/6713178581/job/18244234711?pr=137) with this change)

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,3 +3,6 @@ coverage:
     patch:
       default:
         target: 100%
+    project:
+      default:
+        target: 100%
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -155,6 +155,7 @@ exclude_lines = [
     "if TYPE_CHECKING:",
     "@overload",
     "except ImportError",
+    "return NotImplemented",
     "pass",
 ]
 skip_covered = true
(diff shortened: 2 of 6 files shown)


Carreau:
Thanks, that looks great ! I'll try to find some time to test it with napari to see if it changes anything in the UI.

tlambert03:
i don't know how this got closed!  and now seem unable to reopen.  lemme figure it out

PR #139, opened 2023-11-01
feat: setModelMenuBar accepts instance of MenusRegistry

tlambert03:
This makes it possible to run `setModelMenuBar(app.menus)` to quick generate a menu bar with all menus in an application


tlambert03:
need to consider this:
if you run this on the test app, you get menus: `['File', 'Edit', 'File/Open_From', 'Help']`
the `File/Open_From` is a submenu... so it reveals that we can't just make the menu bar from _all_ menus... we need a way for the user to be involved.  And with that being the case, this becomes a bit less useful, and perhaps should stay as a draft until that issue is considered

PR #142, opened 2023-11-02
docs: documentation overhaul

tlambert03:
general docs overhaul, with better full-api rendering

@lucyleeow, as one who probably looks/works with the docs here most, I'd be grateful for any feedback you have.  I like the general layout better, and it's much more automated, (it does expose a few places where docstrings could be expanded, but happy to follow up on that in another PR)

you can preview the build here: https://app-model--142.org.readthedocs.build/


docs/getting_started.md
```diff
@@ -0,0 +1,246 @@
...
+a dependency injection pattern, provided by the
+[`in-n-out`](https://github.com/pyapp-kit/in-n-out) library.
+
+The application has a [`injection_store`][app_model.Application.injection_store]
+attribute that is an instance of an `in_n_out.Store`.  A `Store` is a collection
+of "providers" (functions that can be called to return an instance of a given
+type), and "processors" (functions that accept an instance of a given type and
+do something with it). The `Store` is used to provide the arguments to the
+command when it is executed, based on the type annotations in the command
+function definition.
```
lucyleeow, line 139: nitpick

```
A `Store` is a collection
of "providers" (functions that can be called to return an instance of a given
type), and "processors" (functions that accept an instance of a given type and
do something with it). The `Store` is used to provide the arguments to the
command when it is executed, based on the type annotations in the command
function definition.
```

Nitpick:
```
A `Store` is a collection of: 

* "providers": functions that can be called to return an instance of a given type. This is used provide the arguments to the command when it is executed, based on the type annotations in the command
function definition.
* "processors": functions that accept an instance of a given type and do something with it. This is used process
   the return value of the command function at execution time, based on command definition return type annotations.
```

lucyleeow: Might feel redundant but a processor example below could be nice.

docs/getting_started.md
```diff
@@ -0,0 +1,246 @@
...
+action = Action(
+    id='greet',
+    title="Greet Current User",
+    callback=print_user_name,
+)
+
+my_app.register_action(action)
+my_app.commands.execute_command('greet')
+# prints "Hi John Doe!"
+```
```
lucyleeow, line 180: Might feel redundant but a processor example below could be nice.

docs/getting_started.md
```diff
@@ -0,0 +1,246 @@
...
+
+    Well, really it's just Qt for now, but the abstraction is valuable for the
+    ability to swap backends.  And we hope to add more backends if the demand is
+    there.
+
+### Qt
+
+Currently, we don't have a generic abstraction for the application window, so
+users are encouraged to directly use the classes in the `app_model.backends.qt`
+module.  One of the main classes is the [`QModelMainWindow`][app_model.backends.qt.QModelMainWindow] object: a subclass of `QMainWindow` that knows how to map
```
lucyleeow, line 198: Biased but might be nice to add a bit about `QModelMenu` ?

lucyleeow (commented):
I started drafting notes for some docs for using app-model in napari and I swear some of it is word for word what you've written here.

I don't fully get contexts but app-model does not keep a registry of contexts? I guess this is why it only updates from an empty context?

https://github.com/pyapp-kit/app-model/blob/9463e678399562c5149e1c4fda75320313db1612/src/app_model/backends/qt/_qaction.py#L171-L175

Maybe a note about this and future plans?

The `getting_started` is really great, logical order and clear explanations (I may be biased because it's similar to how I would have done it :sweat_smile:). I also found the motivations section useful. (Also like that you can easily get links to headers now)

lucyleeow:
Oh while you're here, I noticed that you can't build a `QModelMenu` for submenus because during teardown error results:

see: https://github.com/napari/napari/pull/4865/files#r1288031823

Is this intended?

docs/getting_started.md
tlambert03: good one. thanks added

docs/getting_started.md
tlambert03: #143

docs/getting_started.md
tlambert03: I agree, more of the objects in the qt backend module would benefit from examples.  I think maybe that warrants another full page dedicated to examples using Qt.  Will do in a follow up PR.

tlambert03: #144

tlambert03:
> but app-model does not keep a registry of contexts? I guess this is why it only updates from an empty context?

Not yet.  It's still up to the end-user to manage their contexts. That was always planned, but just something I wanted to consider further.  I do think that a top level `Application.context` would be good and will add and document in a follow up.

> I don't fully get contexts

I think that the [vscode extensions docs](https://code.visualstudio.com/api/references/when-clause-contexts) still provide some of the best general motivation on this... (and of course https://napari.org/stable/guides/contexts_expressions.html) 

but the basic idea is that we need a **declarative** way to express conditions, so that we can set up various things to happen based on those conditions.  In many ways, it overlaps with the concept of signals and slots... but without exposing as much of the underlying business logic to the plugin.   For example, you *could* just literally pass your entire QMainWindow to a plugin and say "go nuts, connect whatever callbacks you want to any of the signals you can find in here"... and then you probably wouldn't need contexts.  But, as you can imagine, that could very easily get unruly, and it also exposes a ton of stuff that you now can no longer easily change without breaking usage.

Contexts let the application say: "here are the things you, the developer, can query about the state of the application".  By declaring special names, like [vscode does here](https://code.visualstudio.com/api/references/when-clause-contexts#available-context-keys) (and like I would encourage napari to do as well, like I started [here](https://napari.org/stable/guides/contexts_expressions.html#napari-context-keys)), you do this in a fully declarative and serializable way.  So the developer can give you an expression `“active_layer_is_rgb | all_layers_same_shape”` using variable names that are present in your context, and *you* (the application) connect the proper signals to make sure that the necessary stuff happens when that expression becomes true/false (for example, making a menu item visible or not, or a button enabled or disabled, or anything like that)

that napari document above was my attempt to capture that in the context of napari, and I'll try to express it here as well but probably in a follow up PR.

tlambert03:
> Oh while you're here, I noticed that you can't build a QModelMenu for submenus because during teardown error results:

exceptions are almost never intended :joy: please open issues when you find stuff like this!  I'm not following napari issues so won't see stuff like that unless you ping me or open an issue.. but I'm more than happy to look into it when you encounter anything unexpected in app-model~

tlambert03:
going to merge this as an improvement over what we already have.  Will follow up with @lucyleeow's improvement suggestions in new PRs

lucyleeow:
> please open issues when you find stuff like this!

Will do thanks. We didn't want to bother you with this one as it wasn't something we needed and we weren't sure of intention. I'll add details to the issue.

tlambert03:
I don't find it a bother :) I like knowing if there's something that the tests aren't catching here!

lucyleeow:
> it would be great to know more specifically what is still confusing about contexts

Thank you for the background, I understood at a high level but the details are useful to know!

I was more confused about the implementation. For some of the `ContextNamespace` and `Context` stuff, I can guess what things are doing but I don't know whats happening 'under the hood'/don't understand the syntax. 

* What does the `['LayerSel']` syntax mean? And the first item `num_selected_layers`, is this like a pydantic field / class attribute?

```python
class LayerListSelectionContextKeys(ContextNamespace['LayerSel']):
    """Available context keys relating to the selection in a LayerList.

    Consists of a default value, a description, and a function to retrieve the
    current value from `layers.selection`.
    """

    num_selected_layers = ContextKey(
        0,
        trans._("Number of currently selected layers."),
        _len,
    )
```

* Where does `get_context` get `_OBJ_TO_CONTEXT`? Is it a global variable?
https://github.com/pyapp-kit/app-model/blob/0140de703687c8132feaaabc7d649ebf325dc5b3/src/app_model/expressions/_context.py#L153

* I saw [this](https://github.com/napari/napari/blob/47b9527ef59bf1b611a6b7f9e3c7877492af3f4a/napari/components/layerlist.py#L81C1-L85C21) note in napari:

```python
        # TODO: figure out how to move this context creation bit.
        # Ideally, the app should be aware of the layerlist, but not vice versa.
        # This could probably be done by having the layerlist emit events that
        # the app connects to, then the `_ctx` object would live on the app,
        # (not here)
```

In what way does the app know about the `layerlist` atm? We use app-models `create_context` but AFAICT we aren't connecting anything to app-model. Or is it more for when app-model has a context registry?

DragaDoncila:
@lucyleeow I have superficial answers to these questions but most likely @tlambert03 has more detail.

> What does the ['LayerSel'] syntax mean? 

The `['LayerSel']` syntax here is, I think based on its type definition [here](https://github.com/napari/napari/blob/main/napari/_app_model/context/_context_keys.py#L11), defining/assigning a generic type for this particular `ContextNamespace`. I'm generally familiar with the idea of a [generic](https://medium.com/@steveYeah/using-generics-in-python-99010e5056eb), but I have to confess I don't really know how it's helpful here... Because the `LayerSel` type isn't a "real" type like `str` or `int` so I don't know what effect it has - is it just for type hinting?

> And the first item num_selected_layers, is this like a pydantic field / class attribute?

I think this is a class attribute, as I don't see any pydantic inheritance in the `Context` chain. Certainly it is accessible as a class attribute on the `LayerListContextKeys` (rather than an instance attribute).

> Where does get_context get _OBJ_TO_CONTEXT? Is it a global variable? (is it updated at runtime?)

`_OBJ_TO_CONTEXT` seems to just be a [module level variable](https://github.com/pyapp-kit/app-model/blob/7f853e606daee0356cc6751e6a6e882c7179724e/src/app_model/expressions/_context.py#L70)? If I understand its purpose, it is to map between a given instance of an object (more specifically, the ID of that object) to the current context of that object, if one has been defined. It is [updated](https://github.com/pyapp-kit/app-model/blob/main/src/app_model/expressions/_context.py#L145) whenever a context is created - though I don't understand the benefit of just storing it on the module like that, rather than as a singleton on the app or something - I guess that's where the context registry would come in?

> In what way does the app know about the layerlist atm?

Right now, I think the app does not know about the layerlist at all. It's napari's job (I think in the qtmainwindow?) to connect the right callbacks to update the context. I think that's the bit that is considered not great and that should be fixed by adding a context registry to app-model...

lucyleeow:
> Because the LayerSel type isn't a "real" type like str or int so I don't know what effect it has - is it just for type hinting?

I guess the question is what is it's purpose? Is it just for type hinting? The other question is a e.g., `LayerListContextKeys` and `LayerListSelectionContextKeys` are both instantiated with generic `Context`s. The only difference is the events they get connected to, so is this just how their different 'input's (layers vs selected layers) are implemented? I guess I did not connect that originally.

tlambert03:
>  Is it just for type hinting?

yep, it's entirely just for type hinting.  whenever you see something like this, where you have a Generic type that is parameterized by some other type, you can think of it as the outer type "taking" the inner type.  Lemme try to explain some of those terms with examples:

```python
from typing import Generic, TypeVar
T = TypeVar("T")

class MyContainer(Generic[T]):
    ...
```

when you see something like this, it means: `MyContainer` is a class that will take, contain, or otherwise deal with some other type.   So, two questions immediately might be:
- What kind of other type?  In this case it *varies* (as indicated by the TypeVar `T`).  As written here, it could be literally anything, but we could have made `T = TypeVar("T",  bound=Union[int, str])`, to state that `MyContainer` only deals with integers or strings
- What does it *do* with that type?  For this you need to see where the type parameter is used... so look deeper into the class:

    ```python
    class MyContainer(Generic[T]):
        def __init__(self, thing: T):
            self._thing = thing
    
        def get_thing(self) -> T:
            return self._thing
    ```

    Ah, ok this is a class that has a `get_thing` method that will return an instance of whatever `T` is... 

Still weird? probably, but the next step is to "parametrize" the generic in a subclass, by replacing the type variable "`T`" with a concrete type:

```python
class MyIntContainer(MyContainer[int]):
   ...
```

The `MyContainer[int]` syntax says that this subclass specifically only deals with integers.  This syntax immediately implies that the `thing` argument to `__init__` *must* be an integer, and it also immediately tells us that the `MyIntContainer .get_thing` method will return an integer.

so, going back to `ContextNamespace`: 

`class LayerListSelectionContextKeys(ContextNamespace['LayerSel']):`

this tells us that `LayerListSelectionContextKeys` is a specific variant of `ContextNamespace` that deals with `LayerSel` types, where `LayerSel` is just an alias for `LayerSel = Selection[Layer]` (i.e. a selection of `Layers`).  Ok, so `LayerListSelectionContextKeys` deals with selection sets of layers... what does it do with them?  We have to look at the `ContextNamespace` code to find out:

```python
class ContextNamespace(Generic[A]):
    def __init__(self, context: MutableMapping) -> None:
        ...
        self._getters: Dict[str, Callable[[A], Any]] = {}  # value getters
```

ok: all `ContextNamespace` subclasses are going to collect a bunch of `getters`, which are defined as functions that take in an instance of the type mentioned above and return anything (`Callable[[A], Any]`).  Meaning, `LayerListSelectionContextKeys` is going to collect a bunch of functions that take a `LayerSel` and do something with them.  (specifically, in this case, it's going to call those functions to update the value of that specific context key when necessary).  Those functions, that take a `LayerSel` are all right here: https://github.com/napari/napari/blob/c6abeb5d6efed2c6ba46e8d415fd3cef5fdf6521/napari/_app_model/context/_layerlist_context.py#L40-L121


----


Why bother with all the type hinting?  because when done properly it can make the developer experience *really* nice by alterting you to possible bugs and errors in your code before you even run anything.  It can tell you that the function you're trying to use as to update one of the keys in your Context is not going to work, because it doesn't accept the right type(s).

see this [bit of the napari docs](https://napari.org/stable/guides/contexts_expressions.html#contextkey-objects) for a bit more on how the type hinting here gives you type feedback in your IDE

-----

is it more trouble than it's worth?  perhaps now it is.  I imagined that I would be working on this in napari for longer, and that doing this with careful type hinting would be able to provide a very nice developer experience.  Now, however, I can see that it does add cognitive burden for those not already intimately familiar with Generic types and parameterization.  You can feel free to either ignore it, or even remove it if it feels too cumbersome.

tlambert03:
btw, i think the mypy docs on generics and parameterization is better than the python docs:
https://mypy.readthedocs.io/en/stable/generics.html

lucyleeow:
:exploding_head: it makes so much more sense now, thanks!

tlambert03:
oh, and there's one more typing-related step:

note how all of the functions `LayerSel` functions (the "getters" whose types were above defined as `Callable[[A], Any]`) return some value.  For example:

```python
def _n_unselected_links(s: LayerSel) -> int:
    from napari.layers.utils._link_layers import get_linked_layers

    return len(get_linked_layers(*s) - s)
```

let's further look to see where the `_n_unselected_links` function is used [in napari](https://github.com/napari/napari/blob/c6abeb5d6efed2c6ba46e8d415fd3cef5fdf6521/napari/_app_model/context/_layerlist_context.py#L187-L191).  Lower down in the file, it's passed as the third argument to `ContextKey`:

```python
    num_unselected_linked_layers = ContextKey(
        0,
        trans._("Number of unselected layers linked to selected layer(s)."),
        _n_unselected_links,
    )
```

well, if you look at the `ContextKey` class itself... you'll see that it too is a generic, that is parametrized by two type variables `A` and `T`.  Specifically, the third argument expects a getter function that takes an `A` and returns a `T` (`Callable[[A], T]`). 

```python
class ContextKey(Name, Generic[A, T]):
    def __init__(
        self,
        default_value: T | __missing = MISSING,
        description: str | None = None,
        getter: Callable[[A], T] | None = None,
        *,
        id: str = "",  # optional because of __set_name__
    ) -> None:
```

So now, we see how the *return* value of that function above might be used in type hinting.  Specifically, it's used further down in the `__get__` method:

```python
class ContextKey(Name, Generic[A, T]):
    def __get__(self, obj: ContextNamespace[A], objtype: type) -> T:
        # When we get from the object, we return the current value
        ...
```

(note that `ContextKey` is a [descriptor object](https://docs.python.org/3/howto/descriptor.html), kinda like `@property` and `__get__` is a special method that determines what will be returned if this name is accessed on an instance of the class in which it is defined as a class variable)

So, if you put that all together:

```python
def _n_unselected_links(s: LayerSel) -> int:
    ...

class LayerListSelectionContextKeys(ContextNamespace['LayerSel']):
    num_unselected_linked_layers = ContextKey(
        0,
        "Number of unselected layers linked to selected layer(s).",
        _n_unselected_links,
    )
```

1. `LayerListSelectionContextKeys` will deal with a set of ContextKeys whose `getter` functions all take a `LayerSel` as  their argument
2. `_n_unselected_links` is a valid `getter` function can take a selection of layers and return an integer (the number of selected layers)
3. `num_unselected_linked_layers` is a context key (i.e. a **name** that has special meaning, which could be used by napari and/or plugin developers to refer declaratively to the concept of "the number of currently selected layers").  Because it's third argument is `_n_unselected_links` (a function that returns an `int`, we know that `num_unselected_linked_layers` will always give the user an integer).

And so if someone were to use `LayerListSelectionContextKeys(). num_unselected_linked_layers ` (either to get or set the value), the IDE would know that it should be an integer, and mypy could give an error if they tried to do otherwise.  And give you hints like this:




----

at least that was the motivation behind the design 🫠

tlambert03:
and (sorry) one more comment:

I recognize that this all is rather complicated.  I want to clarify that I never intended for anyone to have to understand all of that to just *use* the system.  My primary target for napari developers was just this final usage here:

```python
class LayerListSelectionContextKeys(ContextNamespace):
    num_unselected_linked_layers = ContextKey(
        0,
        "Number of unselected layers linked to selected layer(s).",
        _n_unselected_links,
    )
```

I wanted napari developers to be able to use that relatively straightforward syntax to declare context key names, defaults, descriptions, and getter functions – and to be able to group related context keys in a `ContextNamespace` - and have all the rest "just work" in their IDE, giving them type hinting and allowing us to run mypy checks on everything.  I definitely didn't expect more than a couple people to have to follow all of the generics and parametrization.

PR #145, opened 2023-11-03
feat: add top level Application.context

tlambert03:
This PR adds an `Application.context` attribute.  It can serve as a natural place for applications to put a top level context (and, as usual, they can spawn children contexts with `context.new_child`).

it does not yet hook up context.changed signals to anything in the Qt backend.  mostly because I think more consideration of context "scoping" is necessary.  For example, it's not obvious that `app.context` is the "correct" context to pass to one of the `update_from_context` methods... it might be a more narrowly scoped child context provided by the user.


tlambert03:
cc @lucyleeow, this is (in-part) addressing https://github.com/pyapp-kit/app-model/pull/142#pullrequestreview-1712283260

PR #166, opened 2023-12-11
chore: Provide information about callback registered

Czaki:
Provide additional information about command ID collision.


tlambert03 (approved):
thanks!

PR #173, opened 2024-01-07
ci: Add testing of napari/utils/_tests/test_key_bindings.py

Czaki:
Add more test to napari test to easier early find incompatybilites and earlier prepare napari repository for changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ jobs:
       dependency-ref: ${{ matrix.napari-version }}
       dependency-extras: 'testing'
       qt: ${{ matrix.qt }}
-      pytest-args: 'napari/_qt napari/_app_model -k "not async and not qt_dims_2"'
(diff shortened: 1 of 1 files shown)


tlambert03:
thanks @Czaki.  For what it's worth, in this case, I don't think it would have been possible to both satisfy the request I had coming from napari devs in #108 while also keeping all of the tests in napari passing (even with a deprecation warning), since napari itself was asserting the very thing that devs were requesting to have changed

Czaki:
@tlambert03 in a situation like in #108 it will be enough to inform napari-devs that they need to fix tests in napari after such request. 

This allows to not interrupt development workflow and keep working smooth.

tlambert03:
I didn’t know at the time that there were tests outside of app-model that would be affected.  Sorry, it’s probably just gonna happen from time to time that you need to make slight adjustments to your tests

Czaki:
Yes. The next known milestone is [6204](https://github.com/napari/napari/pull/6204) that need to be pushed just after 0.4.19 relese

PR #179, opened 2024-03-17
fix: handle qmods like QT5 for pyside6 not 6.4

psobolewskiPhD:
We're trying to Goldilocks a pyside6 version for napari tests.
Recently (https://github.com/napari/napari/pull/6738) when testing with pyside6 6.3.1, we got this failure:
https://github.com/napari/napari/actions/runs/8303510051/job/22727974608?pr=6738#step:12:209

```
Traceback (most recent call last):
    File "/home/runner/work/napari/napari/napari/_qt/widgets/qt_keyboard_settings.py", line 685, in keyPressEvent
      kb = qkeysequence2modelkeybinding(event_keyseq)
    File "/home/runner/work/napari/napari/.tox/py39-linux-pyside6-cov/lib/python3.9/site-packages/app_model/backends/qt/_qkeymap.py", line 316, in qkeysequence2modelkeybinding
      parts = [SimpleKeyBinding.from_int(qkeycombo2modelkey(x)) for x in iter(key)]
    File "/home/runner/work/napari/napari/.tox/py39-linux-pyside6-cov/lib/python3.9/site-packages/app_model/backends/qt/_qkeymap.py", line 316, in <listcomp>
      parts = [SimpleKeyBinding.from_int(qkeycombo2modelkey(x)) for x in iter(key)]
    File "/home/runner/work/napari/napari/.tox/py39-linux-pyside6-cov/lib/python3.9/site-packages/app_model/backends/qt/_qkeymap.py", line 308, in qkeycombo2modelkey
      qmods = _get_qmods(key)
    File "/home/runner/work/napari/napari/.tox/py39-linux-pyside6-cov/lib/python3.9/site-packages/app_model/backends/qt/_qkeymap.py", line 85, in _get_qmods
      return key.keyboardModifiers()
  AttributeError: 'int' object has no attribute 'keyboardModifiers'
```

In this PR, for versions of pyside6 before 6.4, just use the QT5 handling.
I tested locally with pyside 6.3.1, and latest, as well as pyqt6 6.3.1 and latest.

### DIFF
--- a/src/app_model/backends/qt/_qkeymap.py
+++ b/src/app_model/backends/qt/_qkeymap.py
@@ -3,8 +3,10 @@ from __future__ import annotations
 
 import operator
 from functools import reduce
+
 from typing import TYPE_CHECKING, Mapping, MutableMapping
 
+from qtpy import API, QT_VERSION
 from qtpy.QtCore import QCoreApplication, Qt
 from qtpy.QtGui import QKeySequence
 
@@ -79,7 +81,7 @@ else:
         return int(out)
 
 
-if QT6:
+if QT6 and not (API == "pyside6" and int(QT_VERSION[2]) < 4):
 
     def _get_qmods(key: QKeyCombination) -> Qt.KeyboardModifier:
         return key.keyboardModifiers()
--- a/src/app_model/backends/qt/_qkeymap.py
+++ b/src/app_model/backends/qt/_qkeymap.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 
 import operator
 from functools import reduce
-
 from typing import TYPE_CHECKING, Mapping, MutableMapping
 
 from qtpy import API, QT_VERSION


tlambert03:
works for me.

fwiw, i think the underlying cause here is the fact that the behavior of `QKeySequence.__iter__` (called [here](https://github.com/pyapp-kit/app-model/blob/03f5e1a3fee42a1852935df27a2b50d03bae1107/src/app_model/backends/qt/_qkeymap.py#L316)) is changing.  Another way to fix this would be, in `qkeycombo2modelkey`, to use:

```python
if isinstance(key, int):
    key = QKeyCombination.fromCombined(key)
```

which is a generic (Qt6+) way to cast an integer to a `QKeyCombination`.  but i'll add more qt version testing in a follow up PR and consider a change then.  Thanks @psobolewskiPhD

tlambert03:
@psobolewskiPhD, we're now testing everything from Qt 5.15; 6.2-6.6 https://github.com/pyapp-kit/app-model/pull/180

PR #183, opened 2024-03-23
fix: Do not use lambda in QMenuItemAction.destroyed callback

Czaki:
In napari tests I got segfaults, after local debugs it points to lambda that are modified in this PR. 

In general, using the method is safer, as qt will not try to trigger the method of destroyed objects.

### DIFF
--- a/src/app_model/backends/qt/_qaction.py
+++ b/src/app_model/backends/qt/_qaction.py
@@ -150,9 +150,6 @@ class QMenuItemAction(QCommandRuleAction):
     def _cache_key(app: Application, menu_item: MenuItem) -> tuple[int, int]:
         return (id(app), hash(menu_item))
 
-    def _remove_from_cache(self) -> None:
-        cache_key = QMenuItemAction._cache_key(self._app, self._menu_item)
-        QMenuItemAction._cache.pop(cache_key, None)
 
     @classmethod
     def create(
@@ -180,8 +177,6 @@ class QMenuItemAction(QCommandRuleAction):
             return res
 
         cls._cache[cache_key] = obj = cls(menu_item, app, parent)
-        obj.destroyed.connect(obj._remove_from_cache)
-        app.destroyed.connect(obj._remove_from_cache)
         return obj
 
     def update_from_context(self, ctx: Mapping[str, object]) -> None:
--- a/src/app_model/backends/qt/_qaction.py
+++ b/src/app_model/backends/qt/_qaction.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import contextlib
 from typing import TYPE_CHECKING, ClassVar, Mapping
+from weakref import WeakValueDictionary
 
 from app_model import Application
 from app_model.expressions import Expr
@@ -130,7 +131,7 @@ class QMenuItemAction(QCommandRuleAction):
         Optional parent widget, by default None
     """
 
-    _cache: ClassVar[dict[tuple[int, int], QMenuItemAction]] = {}
+    _cache: ClassVar[dict[tuple[int, int], QMenuItemAction]] = WeakValueDictionary()
 
     def __init__(
         self,
--- a/src/app_model/backends/qt/_qaction.py
+++ b/src/app_model/backends/qt/_qaction.py
@@ -157,12 +157,16 @@ class QMenuItemAction(QCommandRuleAction):
             super().__init__(menu_item.command, app, parent)
             self._menu_item = menu_item
             key = (id(self._app), hash(menu_item))
-            self.destroyed.connect(lambda: QMenuItemAction._cache.pop(key, None))
-            self._app.destroyed.connect(lambda: QMenuItemAction._cache.pop(key, None))
+            self.destroyed.connect(self._remove_from_cache)
+            self._app.destroyed.connect(self._remove_from_cache)
             self._initialized = True
 
         with contextlib.suppress(NameError):
             self.update_from_context(self._app.context)
+
+    def _remove_from_cache(self) -> None:
+        key = (id(self._app), hash(self._menu_item))
+        QMenuItemAction._cache.pop(key, None)
 
     def update_from_context(self, ctx: Mapping[str, object]) -> None:
         """Update the enabled/visible state of this menu item from `ctx`."""
--- a/src/app_model/backends/qt/_qaction.py
+++ b/src/app_model/backends/qt/_qaction.py
@@ -13,6 +13,7 @@ from ._util import to_qicon
 if TYPE_CHECKING:
     from PyQt6.QtGui import QAction
     from qtpy.QtCore import QObject
+    from typing_extensions import Self
 
     from app_model.types import CommandRule, MenuItem
 else:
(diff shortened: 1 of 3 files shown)


Czaki:
I do not understand this test fail

tlambert03:
Thanks for the fix @Czaki.  that test fail was definitely a weird one. Ultimately, I was able to track it down to [this line](https://github.com/pyapp-kit/app-model/pull/183/commits/1fb13fbbcec8b4903ad78f8d6245e9f0e35cf00e#diff-2e1070d7c99e7a6ae71889df8f6ad581161b01565c715e61a9b19adabb807765L320)

```python
action = QMenuItemAction(item, app=app, parent=menu)
```

and a few other things:

1.  The removal of the lambda means that the QMenuItemAction will *not* be removed from the cache when the object is destroyed; because, just like you said qt will not try to trigger the method of destroyed objects. (side note: I'm not 100% sure that this is the behavior we want, but if if the lambda was causing you segfaults, then lets go with it and see if we run into any other unexpected side effects of leaving these items in the cache.). So, this PR made it so that when the menu is cleared during rebuilding, it no longer clears the cache, and that call to `QMenuItemAction()` uses a cached QMenuItemAction instead of making a new one.
2. Oddly enough, it was the providing of `parent=menu` that caused the issue here.  That caused the rebuilding of the menu to hang, so that when the test asserted that the menu items were there, they actually weren't yet there because the menu rebuild never completed.  Exactly *why* it causes the rebuild to hang, I'm not sure... but I suspect it has something to do with doing too much with a single QObject within a callback: in other words, *while* rebuilding a QMenu, if we try to both clear() that menu of all it's QActions, and then add back one of those QActions which never got deleted then stuff goes poorly.

So, simply changing `QMenuItemAction(item, app=app, parent=menu)` to `QMenuItemAction(item, app=app)` was enough to fix it (and I don't think that not passing the parent is a big deal here... since we're about to add it to the menu anyway).

----

you'll note I made a bunch of other changes to the caching pattern in [1fb13fb](https://github.com/pyapp-kit/app-model/pull/183/commits/1fb13fbbcec8b4903ad78f8d6245e9f0e35cf00e).  While trying to debug, I got rid of `__new__` in favor of a `create()` classmethod.  and I think it's probably a safer pattern so stuck with it

tlambert03:
i'll leave this here for the day in case you want to have another look, and then will merge this evening

src/app_model/backends/qt/_qmenu.py
```diff
@@ -305,7 +305,10 @@ def _rebuild(
     exclude: Collection[str] | None = None,
 ) -> None:
     """Rebuild menu by looking up `menu` in `Application`'s menu_registry."""
-    menu.clear()
+    actions = menu.actions()
+    for action in actions:
+        menu.removeAction(action)
```
tlambert03, line 310: interesting. are both of these required to be able to use `create(parent=...)` here?   (i.e. *not* using clear as well as using `setParent()` inside of create?)

happy to go with whatever works best.  just curious

Czaki: We need to keep reference to actions to prevent them being garbage collected, but we need to remove them from the menu before rebuild.

tlambert03: aha... good one

src/app_model/backends/qt/_qaction.py
```diff
@@ -118,52 +120,70 @@ class QMenuItemAction(QCommandRuleAction):
...
         with contextlib.suppress(NameError):
             self.update_from_context(self._app.context)
 
+    @staticmethod
+    def _cache_key(app: Application, menu_item: MenuItem) -> tuple[int, int]:
+        return (id(app), hash(menu_item))
+
+    def _remove_from_cache(self) -> None:
+        cache_key = QMenuItemAction._cache_key(self._app, self._menu_item)
+        QMenuItemAction._cache.pop(cache_key, None)
```
tlambert03, line 155: I'm not too worried about this coverage miss. it's obviously a result of not using a lambda (probably just never gets called in tests anymore).  All good to go otherwise?

Czaki: Hm. When running under debug, it looks that it is really uncalled. 
It looks like `obj.destroyer` is not called because qt do not call self signals on destroyed model. And `app.destroyed` is not called because Qt objects are deleted first. 
So it seems like WeakValueDictionary is enough for self-cleaning.

Czaki, line 155:
```suggestion
```

src/app_model/backends/qt/_qaction.py
```diff
@@ -118,52 +120,70 @@ class QMenuItemAction(QCommandRuleAction):
...
+        app = Application.get_or_create(app) if isinstance(app, str) else app
+        cache_key = QMenuItemAction._cache_key(app, menu_item)
+        if cache_key in cls._cache:
+            res = cls._cache[cache_key]
+            res.setParent(parent)
+            return res
+
+        cls._cache[cache_key] = obj = cls(menu_item, app, parent)
+        obj.destroyed.connect(obj._remove_from_cache)
+        app.destroyed.connect(obj._remove_from_cache)
```
Czaki, line 184:
```suggestion
```

src/app_model/backends/qt/_qaction.py
tlambert03: > It looks like obj.destroyer is not called because qt do not call self signals on destroyed model. And app.destroyed is not called because Qt objects are deleted first.

right, I think it was only the usage of a lambda that was making it get called before this PR.  I assumed that's exactly what you were referring to when you said "as qt will not try to trigger the method of destroyed objects." :joy:

> So it seems like WeakValueDictionary is enough for self-cleaning?

I'm absolutely willing to give it a try

src/app_model/backends/qt/_qaction.py
tlambert03: sure, if you want to!  I agree, they seem kinda useless at this point and perhaps the weak dict is sufficient anyway

tlambert03:
thanks!

Czaki:
Could I request for bug fix release after merge this?

tlambert03:
absolutely.  i prefer to keep releases as close to main as possible and avoid lengthy delays :)

lucyleeow:
Just a question, I did not follow:

> The removal of the lambda means that the QMenuItemAction will not be removed from the cache when the object is destroyed; because, just like you said qt will not try to trigger the method of destroyed objects.

The first part makes sense but I do not understand how the 2nd part. What does "qt will not try to trigger the method of destroyed objects" cause in this context ? Thanks

Czaki:
> The first part makes sense but I do not understand the 2nd part. What does "qt will not try to trigger the method of destroyed objects" cause in this context ? Thanks

Qt (And same for psygnal), when you connect a method to a signal, before calling callback, try first to validate if the object is still alive. And if it is already deleted then callback is not triggered. 

When lamda is used, then there is no option to easy check if all referred objects still exist.

Issue #176, opened 2024-03-05
Allow pass to `QModelMenu` constructor  dict with mapping from types to instances


Czaki:
To avoid depending on injection resolving on menu clicking (see https://github.com/napari/napari/issues/6466) it will be nice to allow providing mapping (like `{Viever: viever_instance}`) to `QModelMenu`. As `QModelMenu`  is often glued to a given window, it will not provide any risk, but reduce error probability. 

The items passed in dict should be stored as weakref internally and unwrapped on trigger fire.

lucyleeow:
@DragaDoncila and I had a look at how this could be implemented but are a bit stumped.

Binding a window to a `QModelMenu` instance is easy enough but it's not clear how best to pass this to an action execution, when the action was executed from a menu.

Ideally we would do this at `execute_command` inside `_on_triggered` [here](https://github.com/pyapp-kit/app-model/blob/a0c33c36dcf9dd2cc9bfd70ab5c4440c74e0f8b2/src/app_model/backends/qt/_qaction.py#L55). But at this step we have no context about where the call came from. I'm also unsure of how best to pass the bound object to `Store.inject`.

A ugly way that could work would be to create a different `QMenuItemAction` in [`_rebuild`](https://github.com/pyapp-kit/app-model/blob/a0c33c36dcf9dd2cc9bfd70ab5c4440c74e0f8b2/src/app_model/backends/qt/_qmenu.py#L300), so that the action that is added to menus, has a callback with the bound argument. I think this would involve making a whole new `CommandRule` though.

Suggestions welcome.

Czaki:
The `QMenuItemAction` 
https://github.com/pyapp-kit/app-model/blob/a0c33c36dcf9dd2cc9bfd70ab5c4440c74e0f8b2/src/app_model/backends/qt/_qaction.py#L118

know its context, as parent menu is passed during creation, 

https://github.com/pyapp-kit/app-model/blob/a0c33c36dcf9dd2cc9bfd70ab5c4440c74e0f8b2/src/app_model/backends/qt/_qmenu.py#L323

so we could reimplement `_on_triggered` in this class.

tlambert03:
i don't yet have strong opinions on this either way, but just for the sake of discussion, can we "motivate" this change a bit more please?

[...]

PR #192, opened 2024-05-03
feat: Allow passing bound parameters `QModelMenu`

lucyleeow:
WIP: appreciate input on whether I'm on the right track.

Attempt at #176

Allow 'binding' of objects to `QModelMenu` that will be passed to action callbacks if the type annotation matches (inspired by `Store.inject` from `in-n-out`).
Reimplements `_on_triggered` of `QMenuItemAction` as suggested by @Czaki.

Bound objects passed as kwarg to callback via `CommandsRegistry.execute_command` `kwargs` param. The problem with this though is that these kwargs are passed after we run `run_injected`:

https://github.com/pyapp-kit/app-model/blob/a0c33c36dcf9dd2cc9bfd70ab5c4440c74e0f8b2/src/app_model/registries/_commands_reg.py#L181-L192

 In `run_injected` we give precedence to caller provided args/kwargs but AFAICT this cannot be done with kwargs passed via `execute_command`. Also I think (?) we have the problem that `execute_command(kwarg)` could be providing a kwarg that was already provided by `run_injected`? I am not sure how this parameter was meant to be used originally @tlambert03 ?

Edit: TODO I think I'll want to check if the bound obj is a weakref.

cc @DragaDoncila

### DIFF
--- a/src/app_model/backends/qt/_qaction.py
+++ b/src/app_model/backends/qt/_qaction.py
@@ -154,8 +154,8 @@ class QMenuItemAction(QCommandRuleAction):
         sig = _resolve_sig_or_inform(
             callback,
             localns=self._app._injection_store.namespace,
-            on_unresolved_required_args=False,
-            on_unannotated_required_args=False,
+            on_unresolved_required_args="ignore",
+            on_unannotated_required_args="ignore",
             guess_self=False,
         )
         if sig is None:
--- a/src/app_model/backends/qt/_qaction.py
+++ b/src/app_model/backends/qt/_qaction.py
@@ -4,6 +4,8 @@ import contextlib
 from typing import TYPE_CHECKING, ClassVar, Mapping
 from weakref import WeakValueDictionary
 
+from in_n_out._type_resolution import _resolve_sig_or_inform
+
 from app_model import Application
 from app_model.expressions import Expr
 from app_model.types import ToggleRule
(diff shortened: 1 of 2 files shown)


src/app_model/backends/qt/_qaction.py
```diff
@@ -146,6 +148,29 @@ def __init__(
         with contextlib.suppress(NameError):
             self.update_from_context(self._app.context)
 
+    def _on_triggered(self, checked: bool) -> None:
+        # Pass bound `QModelMenu` objects when callback parameter annotation matches
+        callback = self._app.commands[self._command_id].resolved_callback
+        sig = _resolve_sig_or_inform(
```
tlambert03, line 154: this really isn't something that should be used at callback time, it's expensive and complicated.  the goal of this function in in-n-out was to make sure, *at injection time* that we had everything we needed to later match up the types to providers.  But I don't like the idea of resolution being done every time a command is called.

tlambert03 (commented):
to be honest, I'm not thrilled by this approach.  I'm not strongly convicted either way yet, but this has a bit of a code smell.  Notably, until this PR, the only thing we've ever needed from in-n-out here is just a single `Store` import.  Now, we're importing a deep private function that (in-n-out uses only once at injection time), and calling it _every_ time a command is called.  So, it kinda stands out as a distinct departure from the other existing patterns.  For that reason, I'd like a bit more motivation for the root problem, and what this is the best way to fix it (and not just "a way" to fix it)

lucyleeow:
I'm going to close this as I agree, I don't think this implementation would be the way to do things and I just don't see a way where it wouldn't impact performance.

We could get the parent widget from `QModelMenu` a different way, e.g., via a provider.

tlambert03:
Thanks for your work on this @lucyleeow

PR #194, opened 2024-05-06
feat: give registries more control over registration of actions

tlambert03:
This PR moves some of the logic in the `register_action` function into methods on the registries themselves.  This will make it easier for custom subclasses of registries to control exactly how they register actions.  
It has some relatively inconsequential other stuff, like making `RegisteredCommand` immutable and public

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,14 +15,14 @@ concurrency:
 
 jobs:
   test:
-    uses: pyapp-kit/workflows/.github/workflows/test-pyrepo.yml@v1
+    uses: pyapp-kit/workflows/.github/workflows/test-pyrepo.yml@v2
     secrets: inherit
     with:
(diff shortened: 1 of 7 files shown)


tlambert03:
@Czaki, with this PR, you should be able to use your own `Action` subclass.

1. do whatever you want to subclass and extend `Action`
2. do whatever you want to subclass and extend `CommandRegistry` (note that there is now a `CommandRegistry.register_action` method that you can extend/override)
3. pass your `CommandRegistry` subclass type [here](https://github.com/napari/napari/blob/1dcbadea20149c43d49764afd3fe8b39b76998a9/napari/_app_model/_app.py#L26C9-L26C70) when you initialize your app  (e.g. `super().__init__(app_name, raise_synchronous_exceptions=True, commands_reg_class=NapariCommandRegistry)`

Issue #196, opened 2024-05-16
`enablement`/`when` typing strictly `Expr` ?


lucyleeow:
### Description

Is the typing for `enablement` and `when` strictly only `Expr` ?
https://github.com/pyapp-kit/app-model/blob/8289d03f3b4fb73f886e36bc53132cc48b32d941/src/app_model/types/_command_rule.py#L66

I thought the typing would be:

https://github.com/pyapp-kit/app-model/blob/8289d03f3b4fb73f886e36bc53132cc48b32d941/src/app_model/expressions/_expressions.py#L81

Indeed I see in app-model tests a string is used:

https://github.com/pyapp-kit/app-model/blob/8289d03f3b4fb73f886e36bc53132cc48b32d941/tests/conftest.py#L140

lucyleeow:
ref: https://github.com/napari/napari/pull/6864/files#r1602701616

lucyleeow:
@tlambert03 slightly related question, how do you think it should be handled when `enablement` is a function call, that is too complicated to be made into a supported `Expression`?

In #6864 we've set a context key to be the result of that function call, and we do this within `show_context_menu` which I think is connected to aboutToShow. I don't think this is the blessed way and it seems hacky?

cc @dragadoncila

tlambert03:
I commented in https://github.com/napari/napari/pull/6864#discussion_r1603003381

Point is: the type annotations declares what it will be after pydantic performance type coercion.  If you give a string, it will turn it into an Expr.  If you want to avoid type errors in mypy, you can add the pydantic mypy plugin: https://docs.pydantic.dev/latest/integrations/mypy/

lucyleeow:
Ah right, good to know. Will close but if you have thoughts on how best to handle function call 'expressions', please let me know!

tlambert03:
> @tlambert03 slightly related question, how do you think it should be handled when enablement is a function call, that is too complicated to be made into a supported Expression?  In https://github.com/napari/napari/pull/6864 we've set a context key to be the result of that function call, and we do this within show_context_menu which I think is connected to aboutToShow. I don't think this is the blessed way and it seems hacky?

I think "too complicated to be made into a supported Expression" is probably the thing to dig into here.

First, enablement is *always* a function call in a sense.  Specifically, enablement is the result of evaluating the expression, substituting in the values of the context for variable names in the expression:

```python
>>> expression = 'x + 7 < z'
>>> eval(expression, {'x': 5, 'z': 15})
[...]

PR #198, opened 2024-05-16
feat: allow callable values in contexts

tlambert03:
This implements the idea I was mentioning in [#196 ](https://github.com/pyapp-kit/app-model/issues/196#issuecomment-2114978480), wherein a context *value* is allowed to be a callable (that takes no arguments).

it seems the performance penalty isn't too bad (but of course it would depend on how slow the function is that has to be evaluated to fill in the context:

```python
import timeit

from app_model.expressions import parse_expression

expr = parse_expression("x + 7 < z")


for rn in (1, 2, 3, 4):
    ctx = {str(k): k for k in range(10**rn)}
    ctx["x"] = 3
    ctx["z"] = 10

    N = 100000

    a = timeit.timeit("expr.eval(ctx)", number=N, globals=globals())
    expr.eval(ctx)

    ctx["x"] = lambda: 3
    ctx["z"] = lambda: 10
    b = timeit.timeit("expr.eval_callable_context(ctx)", number=N, globals=globals())

    print(f"with {len(ctx)} variables...")
    print(f"  a is {b/a:.2f} times faster than b")
```

```
with 12 variables...
  a is 1.05 times faster than b
with 102 variables...
  a is 1.04 times faster than b
with 1002 variables...
  a is 1.05 times faster than b
with 10002 variables...
  a is 1.06 times faster than b
```

### DIFF
--- a/src/app_model/expressions/_expressions.py
+++ b/src/app_model/expressions/_expressions.py
@@ -177,6 +177,7 @@ class Expr(ast.AST, Generic[T]):
             raise RuntimeError("Don't instantiate Expr. Use `Expr.parse`")
         super().__init__(*args, **kwargs)
         ast.fix_missing_locations(self)
+        self._names = set(_iter_names(self))
 
     def eval(self, context: Mapping[str, object] | None = None) -> T:
         """Evaluate this expression with names in `context`."""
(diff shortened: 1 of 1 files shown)


tlambert03:
@lucyleeow @DragaDoncila @Czaki  ... let me know if you'd like this feature:

essentially, it would allow you to undo the changes made to layer_delegate: https://github.com/napari/napari/blob/4a56237e94480934b1b9f5df5c63949a4059b1f8/napari/_qt/containers/_layer_delegate.py#L325-L328

and instead "permanently" set `ctx['valid_spatial_json_clipboard'] = is_valid_spatial_in_clipboard` (rather than doing it every time the menu is clicked).  And if (and only if) the expression being evaluated needed `valid_spatial_json_clipboard` would it be evaluated.  The downside would be that it would be evaluated for *each* expression... so if you know that there are a number of expressions that need that value, it will be slower than evaluating it once for all of them.  

However, in any case, I still would recommend not calculating `is_valid_spatial_in_clipboard` every time the layer menu is opened, but rather only once on `QClipboard.changed`, as mentioned in https://github.com/pyapp-kit/app-model/issues/196#issuecomment-2114978480

tlambert03:
moving to #199 ... accidentally had this on my main branch

Issue #210, opened 2024-07-08
Consider changing the default keybinding separator on printout to `-`


jni:
* app-model version: 0.2.7
* Python version: 3.11
* Operating System: macOS

### Description

Typically when I think about settings, config, etc, I try to make things as easily human-editable as possible. This one is minor but `Control` is easier to type than `Ctrl` (#209), and `Ctrl-Y` is easier to type than `Ctrl+Y`. (I also find it marginally easier to read.)

The keybindings modifiers [parser](https://github.com/pyapp-kit/app-model/blob/e527101f56b16df80dd02eadf0d6a4075f8468e8/src/app_model/types/_keys/_keybindings.py#L254-L294) already allows `-` when reading, so it would not be a big change (I hope) to change what is written down.

tlambert03:
i think this should be an app-specific preference at best.  using `+` is very common, and appeared to be the more common delimiter when I looked around when first implementing this (I think it has the advantage of semantics "this key plus this key").  but i do think you should be able to show them however you want to, and perhaps #211 would be a good place to do that

PR #211, opened 2024-07-10
feat: add a way to get a user-facing string representation of keybindings

### LINKED ISSUE #89: idea: add user-facing rendering to keybindings
in napari, we use a system to render shortcuts with special characters (e.g. the Mac Command symbol for Cmd). I’m on my phone right now so it’s hard for me to show what I mean but I hope you understand the gist

I was thinking that this sort of almost “text rendering” should belong in app-model since I imagine it has a pretty common use case

what do you think @tlambert03? I’d like to work on this but wanted to check first if people like the idea

dalthviz:
Hi there, this is an initial attempt to add a way to get a user-facing string representation of keybindings (with optional key symbols and taking into account OS differences) by porting the logic already implemented over napari (https://github.com/napari/napari/pull/5064 and https://github.com/napari/napari/pull/5604).

This is still a work in progress but any early feedback is appreciated!

Closes #89

### DIFF
--- a/src/app_model/types/_keys/_key_codes.py
+++ b/src/app_model/types/_keys/_key_codes.py
@@ -154,13 +154,32 @@ class KeyCode(IntEnum):
     PauseBreak = auto()
 
     def __str__(self) -> str:
+        """Get a normalized string representation (constant to all OSes) of this `KeyCode`."""
         return keycode_to_string(self)
 
-    def to_os_symbol(self, os: Optional[OperatingSystem] = None) -> str:
+    def os_symbol(self, os: Optional[OperatingSystem] = None) -> str:
+        """Get a string representation of this `KeyCode` using a symbol/OS specific symbol.
+
+        Some examples:
+            * `KeyCode.Enter` is represented by `↵`
+            * `KeyCode.Meta` is represented by `⊞` on Windows, `Super` on Linux and `⌘` on MacOS
+
+        If no OS is given, the current detected one is used.
+        """
         os = OperatingSystem.current() if os is None else os
         return keycode_to_os_symbol(self, os)
 
-    def to_os_name(self, os: Optional[OperatingSystem] = None) -> str:
+    def os_name(self, os: Optional[OperatingSystem] = None) -> str:
+        """Get a string representation of this `KeyCode` using the OS specific naming for the key.
+
+        This differs from `__str__` since with it a normalized representation (constant to all OSes) is given.
+        Sometimes these representations coincide but not always! Some examples:
+            * `KeyCode.Enter` is represented by `Enter` (`__str__` represents it as `Enter`)
+            * `KeyCode.Meta` is represented by `Win` on Windows, `Super` on Linux and `Cmd` on MacOS
+            (`__str__` represents it as `Meta`)
+
+        If no OS is given, the current detected one is used.
+        """
         os = OperatingSystem.current() if os is None else os
         return keycode_to_os_name(self, os)
 
--- a/src/app_model/types/_keys/_key_codes.py
+++ b/src/app_model/types/_keys/_key_codes.py
@@ -6,12 +6,15 @@ from typing import (
     Dict,
     Generator,
     NamedTuple,
+    Optional,
     Set,
     Tuple,
     Type,
     Union,
     overload,
 )
+
+from app_model.types._constants import OperatingSystem
 
 if TYPE_CHECKING:
     from pydantic.annotated_handlers import GetCoreSchemaHandler
@@ -24,6 +27,7 @@ __all__ = ["KeyCode", "KeyMod", "ScanCode", "KeyChord"]
 
 # flake8: noqa
 # fmt: off
+
 
 class KeyCode(IntEnum):
     """Virtual Key Codes, the integer value does not hold any inherent meaning.
@@ -151,6 +155,14 @@ class KeyCode(IntEnum):
 
     def __str__(self) -> str:
         return keycode_to_string(self)
+
+    def to_os_symbol(self, os: Optional[OperatingSystem] = None) -> str:
+        os = OperatingSystem.current() if os is None else os
+        return keycode_to_os_symbol(self, os)
+
+    def to_os_name(self, os: Optional[OperatingSystem] = None) -> str:
+        os = OperatingSystem.current() if os is None else os
+        return keycode_to_os_name(self, os)
 
     @classmethod
     def from_string(cls, string: str) -> 'KeyCode':
@@ -429,6 +441,8 @@ _NATIVE_WINDOWS_VK_TO_KEYCODE: Dict[str, KeyCode] = {}
 def _build_maps() -> Tuple[
     Callable[[KeyCode], str],
     Callable[[str], KeyCode],
+    Callable[[KeyCode, OperatingSystem], str],
+    Callable[[KeyCode, OperatingSystem], str],
     Callable[[ScanCode], str],
     Callable[[str], ScanCode],
 ]:
@@ -571,6 +585,48 @@ def _build_maps() -> Tuple[
         'cmd': KeyCode.Meta,
     }
 
+    # key symbols mappings per platform
+    WIN_KEY_SYMBOLS: dict[str, str] = {
+        "Ctrl": "Ctrl",
+        "Shift": "⇧",
+        "Alt": "Alt",
+        "Meta": "⊞",
+        "Left": "←",
+        "Right": "→",
+        "Up": "↑",
+        "Down": "↓",
+        "Backspace": "⌫",
+        "Delete": "⌦",
+        "Tab": "↹",
+        "Escape": "Esc",
+        "Return": "⏎",
+        "Enter": "↵",
+        "Space": "␣",
+    }
+    MACOS_KEY_SYMBOLS: dict[str, str] = {**WIN_KEY_SYMBOLS, "Ctrl": "⌃", "Alt": "⌥", "Meta": "⌘"}
+    LINUX_KEY_SYMBOLS: dict[str, str] = {**WIN_KEY_SYMBOLS, "Meta": "Super"}
+
+    # key names mappings per platform
+    WIN_KEY_NAMES: dict[str, str] = {
+        "Ctrl": "Ctrl",
+        "Shift": "Shift",
+        "Alt": "Alt",
+        "Meta": "Win",
+        "Left": "Left",
+        "Right": "Right",
+        "Up": "Up",
+        "Down": "Down",
+        "Backspace": "Backspace",
+        "Delete": "Supr",
+        "Tab": "Tab",
+        "Escape": "Esc",
+        "Return": "Return",
+        "Enter": "Enter",
+        "Space": "Space",
+    }
+    MACOS_KEY_NAMES: dict[str, str] = {**WIN_KEY_NAMES, "Ctrl": "Control", "Alt": "Option", "Meta": "Cmd"}
+    LINUX_KEY_NAMES: dict[str, str] = {**WIN_KEY_NAMES, "Meta": "Super"}
+
     seen_scancodes: Set[ScanCode] = set()
     seen_keycodes: Set[KeyCode] = set()
     for i, km in enumerate(_MAPPINGS):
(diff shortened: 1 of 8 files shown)


src/app_model/types/_keys/_keybindings.py
```diff
@@ -19,6 +19,58 @@
 _re_cmd = re.compile(r"cmd[\+|\-]")
 
 
+def get_keys_user_facing_representation(
```
tlambert03, line 22: probably don't need a function to re-define these dicts every time a string is generated, since they won't change at all at runtime, the only thing is which dict to pull from... so maybe a module level:

```python
WIN_KEY_SYMBOLS: dict[str, str] = {...}
MACOS_KEY_SYMBOLS: dict[str, str] = {**WIN_SYMBOLS, "Ctrl": "⌃", "Alt": "⌥", "Meta": "⌘"}
LINUX_KEY_SYMBOLS: dict[str, str] = {**WIN_SYMBOLS, "Meta": "Super"}

# same for names
```

and then wherever you would have previously used this function, just pick a dict

src/app_model/types/_keys/_keybindings.py
```diff
@@ -117,6 +169,33 @@ def to_int(self, os: Optional[OperatingSystem] = None) -> int:
...
+
+    def to_text(
+        self, os: Optional[OperatingSystem] = None, use_symbols: bool = False
+    ) -> str:
+        """Get a user-facing string representation of this SimpleKeyBinding.
+
+        Optionally, the string representation can be constructed with symbols
+        like ↵ for Enter or OS specific ones like ⌘ for Cmd on MacOS.
+        """
+        joinchar, key_symbols, key_names = get_keys_user_facing_representation(os=os)
```
tlambert03, line 193: i think @jni wants more control over the joinchar (see #210) so joinchar should be a parameter with a default fallback if not provided

src/app_model/types/_keys/_keybindings.py
```diff
@@ -117,6 +169,33 @@ def to_int(self, os: Optional[OperatingSystem] = None) -> int:
...
+        return joinchar.join(
+            key_symbols.get(x, x) if use_symbols else key_names.get(x, x)
```
tlambert03, line 195: i wonder if much of this PR (including the look into the mapping here with `.get(x, x)`) should actually be a method on `KeyCode`.  Currently, this method is first casting `self.key` (which is a `KeyCode`) to a string, and then using that string to again index into another map.  What if did that more directly:

```python
class KeyCode(IntEnum):
    ...
    def to_os_symbol(self, os: Optional[OperatingSystem]) -> str: ...
    def to_os_name(self, os: Optional[OperatingSystem]) -> str: ...
```

and then put these lookups at a lower level in `_key_codes.py`?

that would also have the benefit of being able to directly get these symbols at a lower level, given something like `KeyCode.Meta` ... without first having to put it into a keybinding object

tlambert03 (commented):
thanks @dalthviz, this would be a nice addition.  couple thoughts about where this should live, before we consider the final api

src/app_model/types/_keys/_keybindings.py
tlambert03: (however, see later comment)

src/app_model/types/_keys/_keybindings.py
dalthviz: Oh I see, I think that makes sense, will give it a try :+1: also thank you so much for the early feedback!

dalthviz: Thank you! And actually, checking `_key_codes.py`, I'm wondering if it could make sense to add the symbols and names mappings to the logic over `_build_maps` along side the addition there of some new functions like:

```python
def _keycode_to_os_symbol(keystr: str, os: OperatingSystem) -> str:...
def _keycode_to_os_name(keystr: str, os: OperatingSystem) -> str:...
```

Seems like that could be a good place 🤔 What do you think?

tlambert03: yeah i agree, `_build_maps` is the right place to construct these lookups.

dalthviz: Pushed some changes which hopefully should improve code organization :+1: Also, checking https://github.com/napari/napari/pull/5064 I think another functionality that could be added is https://github.com/napari/napari/blob/main/napari/utils/interactions.py#L312 so not completly 1 to 1 but kind of a `from_text` method. Does that makes sense?

src/app_model/types/_keys/_key_codes.py
```diff
@@ -571,6 +585,48 @@ class _KM(NamedTuple):
...
+        "Backspace": "Backspace",
+        "Delete": "Supr",
+        "Tab": "Tab",
+        "Escape": "Esc",
+        "Return": "Return",
+        "Enter": "Enter",
+        "Space": "Space",
+    }
+    MACOS_KEY_NAMES: dict[str, str] = {**WIN_KEY_NAMES, "Ctrl": "Control", "Alt": "Option", "Meta": "Cmd"}
+    LINUX_KEY_NAMES: dict[str, str] = {**WIN_KEY_NAMES, "Meta": "Super"}
```
tlambert03, line 628: definitely nit-picky, but we can save ourselves the `if os ==` conditionals on lookup by adding on more level to these maps... (then see associated change below)

```suggestion
    # key symbols on all platforms
    KEY_SYMBOLS: dict[KeyCode, str] = {
        KeyCode.Shift: "⇧",
        KeyCode.LeftArrow: "←",
        KeyCode.RightArrow: "→",
        KeyCode.UpArrow: "↑",
        KeyCode.DownArrow: "↓",
        KeyCode.Backspace: "⌫",
        KeyCode.Delete: "⌦",
        KeyCode.Tab: "⇥",
        KeyCode.Escape: "⎋",
        KeyCode.Enter: "↵",
        KeyCode.Space: "␣",
        KeyCode.CapsLock: "⇪",
    }
    # key symbols mappings per platform
    OS_KEY_SYMBOLS: dict[OperatingSystem, dict[KeyCode, str]] = {
        OperatingSystem.WINDOWS: {**KEY_SYMBOLS, KeyCode.Meta: "⊞"},
        OperatingSystem.LINUX: {**KEY_SYMBOLS, KeyCode.Meta: "Super"},
        OperatingSystem.MACOS: {
            **KEY_SYMBOLS,
            KeyCode.Ctrl: "⌃",
            KeyCode.Alt: "⌥",
            KeyCode.Meta: "⌘",
        },
    }

    # key names mappings per platform
    OS_KEY_NAMES: dict[OperatingSystem, dict[KeyCode, str]] = {
        OperatingSystem.WINDOWS: {KeyCode.Meta: "Win"},
        OperatingSystem.LINUX: {KeyCode.Meta: "Super"},
        OperatingSystem.MACOS: {
            KeyCode.Ctrl: "Control",
            KeyCode.Alt: "Option",
            KeyCode.Meta: "Cmd",
        },
    }
```

src/app_model/types/_keys/_key_codes.py
```diff
@@ -602,6 +658,22 @@ def _keycode_from_string(keystr: str) -> KeyCode:
...
+            return LINUX_KEY_SYMBOLS.get(str(keycode), str(keycode))
+        return WIN_KEY_SYMBOLS.get(str(keycode), str(keycode))
+
+    def _keycode_to_os_name(keycode: KeyCode, os: OperatingSystem) -> str:
+        """Return key name for an OS for a given KeyCode."""
+        if os == OperatingSystem.MACOS:
+            return MACOS_KEY_NAMES.get(str(keycode), str(keycode))
+        elif os == OperatingSystem.LINUX:
+            return LINUX_KEY_NAMES.get(str(keycode), str(keycode))
+        return WIN_KEY_NAMES.get(str(keycode), str(keycode))
```
tlambert03, line 675:
```suggestion
    def _keycode_to_os_symbol(keycode: KeyCode, os: OperatingSystem) -> str:
        """Return key symbol for an OS for a given KeyCode."""
        if keycode in (symbols := OS_KEY_SYMBOLS.get(os, {})):
            return symbols[keycode]
        return str(keycode)

    def _keycode_to_os_name(keycode: KeyCode, os: OperatingSystem) -> str:
        """Return key name for an OS for a given KeyCode."""
        if keycode in (names := OS_KEY_NAMES.get(os, {})):
            return names[keycode]
        return str(keycode)
```

note, this also avoids the additional lookup triggered by `str(keycode)` until absolutely necessary

src/app_model/types/_keys/_key_codes.py
```diff
@@ -152,6 +156,14 @@ class KeyCode(IntEnum):
     def __str__(self) -> str:
         return keycode_to_string(self)
 
+    def to_os_symbol(self, os: Optional[OperatingSystem] = None) -> str:
```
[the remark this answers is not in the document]
tlambert03: also please add docstrings to these two public methods, with a brief clarification on how `to_name` differs from `__str__`

src/app_model/types/_keys/_keybindings.py
```diff
@@ -210,6 +246,24 @@ def to_int(self, os: Optional[OperatingSystem] = None) -> int:
             return KeyChord(*parts)
         return parts[0]
 
+    def to_text(
+        self,
+        os: Optional[OperatingSystem] = None,
+        use_symbols: bool = False,
+        joinchar: str = "+",
+    ) -> str:
+        """Get a text representation of this KeyBinding.
```
tlambert03, line 255: i think we need a slight additional clarification on when this would be used with `symbols==False` instead of just using `__str__`.  Currently, it reads like the os-specificity is *only* for the symbols. (same with other docstring)

And, since we're adding this, i think we should perhaps also add a docstring to `__str__` clarifying that it returns a normalized string that is constant across all OSes

src/app_model/types/_keys/_keybindings.py
```diff
@@ -110,6 +110,42 @@ def to_int(self, os: Optional[OperatingSystem] = None) -> int:
...
+        """Create KeyCode instances list of modifiers from this SimpleKeyBinding."""
+        mods = []
+        if self.ctrl:
+            mods.append(KeyCode.from_string("Ctrl"))
+        if self.shift:
+            mods.append(KeyCode.from_string("Shift"))
+        if self.alt:
+            mods.append(KeyCode.from_string("Alt"))
+        if self.meta:
+            mods.append(KeyCode.from_string("Meta"))
```
tlambert03, line 123:
```suggestion
            mods.append(KeyCode.Ctrl)
        if self.shift:
            mods.append(KeyCode.Shift)
        if self.alt:
            mods.append(KeyCode.Alt)
        if self.meta:
            mods.append(KeyCode.Meta)
```

tlambert03 (commented):
getting close!  i made some refactoring suggestions, and some other minor requests

src/app_model/types/_keys/_keybindings.py
tlambert03: i believe #213 should take care of that?  but let me know if I'm not understanding what that code does

dalthviz: Give that a check and I think the work at #213 partialy captures the `parse_platform` logic at  https://github.com/napari/napari/blob/main/napari/utils/interactions.py#L312 The missing thing I would say is also being able to parse symbols from non modifier keys like enter, arrows, backspace, delete, tab, space, etc. Could it be worthy to further extend the parsing capabilities to any symbol or representation that `os_symbol` and `os_name` can return? From a quick check, maybe extending the `KEYCODE_FROM_LOWERCASE_STRING` dictionary could be a way to achieve this? So something like:

```python
KEYCODE_FROM_LOWERCASE_STRING: Dict[str, KeyCode] = {
    # two special cases for assigning os-specific strings to the meta key
    'win': KeyCode.Meta,
    'cmd': KeyCode.Meta,
    # special cases for symbols and other os-specific strings
    '↵': KeyCode.Enter,
    '→': KeyCode.RightArrow,
    ...
}
```

tlambert03: ah I see... yeah, perhaps.  Lets discuss that in another issue though.  Seems like it might be a cool but YAGNI feature?  do you actively parse strings with key symbols in napari?  or was it more of a "why not allow this" sort of feature?

dalthviz: Created https://github.com/pyapp-kit/app-model/issues/214 :+1: and I think that was added to be able to store and comper shortcuts in a canonical/normalized form from a representation of a keybinding like the one you would get from the GUI (like for example the napari shortcuts widget) 🤔 So, to support logic like the following: 

https://github.com/napari/napari/blob/5e685a20c9563a42576ead39a4f0286337a52ce6/napari/_qt/widgets/qt_keyboard_settings.py#L403-L428

tlambert03: Ok, plenty to discuss there.  I would definitely keep things as KeyCodes and not strings as much as possible.  So, rather than using current.text() in that widget, you should take advantage of itemData(), or somehow have the visual representation of the keybinding in the widget not _be_ the actual data that napari uses in the model.  (Think of this as a classic model view thing, where the model is the KeyCode, and the view is some string that the user sees)

dalthviz: Yep, that makes sense to me :+1: I think I will mark this as ready for review then. Let me know if further changes are needed here!

tlambert03 (commented):
thank you @dalthviz, lovely

PR #212, opened 2024-07-11
feat: Add `filter_keybinding` to `KeyBindingRegistry`

### LINKED ISSUE #206: Allow 'filter' option in keybinding registry
Would you consider having a 'filter' option in the keybinding registry to allow users to be more strict on what is allowed as a key sequence?

app-model is pretty open on what is considered a valid, but users may want to:

* not allow certain `KeyCode`s
* not allow certain `KeyCombo`s - e.g., modifier only key combos, which you were also against supporting (in napari) here: https://github.com/napari/napari/issues/5747#issuecomment-1515301910, 
   * also app-model allows multiple modifiers (e.g., 'alt+alt', 'alt+alt+alt'),  which we may (?) want to not support. I can't quite tell if this is problematic only with napari legacy code but Kira did implement a function to de-duplicate (I do know that these two result in the same int, though `alt` by itself is a different int)

I'm also okay with implementing downstream if you don't think it belongs in app-model

cc @dragadoncila

lucyleeow:
closes #206

Adds a `filter_keybinding` (not 100% on name) attr to `KeyBindingRegistry`, which takes `Callable[[KeyBinding], str]` - if str is empty the keybinding is valid otherwise error message str should be returned.

Just a start, happy to take any feedback.

Will update docs once code ready.

### DIFF
--- a/src/app_model/registries/_keybindings_reg.py
+++ b/src/app_model/registries/_keybindings_reg.py
@@ -31,8 +31,8 @@ class KeyBindingsRegistry:
     Attributes
     ----------
     filter_keybinding : Callable[[KeyBinding], str] | None
-        Optional function for filtering out invalid `KeyBinding`s.
-        Callable should return a error message string if `KeyBinding` is invalid
+        Function for filtering out invalid `KeyBinding`s.
+        Callable should return an error message string if `KeyBinding` is invalid
         or empty string if `KeyBinding` is valid.
     """
 
--- a/src/app_model/registries/_keybindings_reg.py
+++ b/src/app_model/registries/_keybindings_reg.py
@@ -86,11 +86,10 @@ class KeyBindingsRegistry:
                 _keyb = keyb
 
             try:
-                d = self.register_keybinding_rule(action.id, _keyb)
+                if d := self.register_keybinding_rule(action.id, _keyb):
+                    disposers.append(d)
             except ValueError as e:
                 msg.append(str(e))
-            else:
-                disposers.append(d)
         if msg:
             raise ValueError(
                 "The following keybindings were not valid:\n" + "\n".join(msg)
--- a/src/app_model/registries/_keybindings_reg.py
+++ b/src/app_model/registries/_keybindings_reg.py
@@ -31,9 +31,9 @@ class KeyBindingsRegistry:
     Attributes
     ----------
     filter_keybinding : Callable[[KeyBinding], str] | None
-        Function for filtering out invalid `KeyBinding`s.
-        Callable should return an error message string if `KeyBinding` is invalid
-        or empty string if `KeyBinding` is valid.
+        Optional function for applying additional `KeyBinding` filtering.
+        Callable should accept a `KeyBinding` object and return an error message
+        (`str`) if `KeyBinding` is rejected, or empty string otherwise.
     """
 
     registered = Signal()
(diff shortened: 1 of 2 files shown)


src/app_model/registries/_keybindings_reg.py
```diff
@@ -26,12 +26,30 @@ class _RegisteredKeyBinding(NamedTuple):
 
 
 class KeyBindingsRegistry:
-    """Registry for keybindings."""
+    """Registry for keybindings.
+
+    Attributes
+    ----------
+    filter_keybinding : Callable[[KeyBinding], str] | None
```
lucyleeow, line 33: Is this where attr docs should be?

src/app_model/registries/_keybindings_reg.py
```diff
@@ -26,12 +26,30 @@ class _RegisteredKeyBinding(NamedTuple):
...
+        self._filter_keybinding: Callable[[KeyBinding], str] | None = None
+
+    @property
+    def filter_keybinding(self) -> Callable[[KeyBinding], str] | None:
+        """Return the `filter_keybinding`."""
+        return self._filter_keybinding
+
+    @filter_keybinding.setter
+    def filter_keybinding(self, filter: Callable[[KeyBinding], str] | None) -> None:
+        self._filter_keybinding = filter
```
lucyleeow, line 52: Should we be doing any type checking?

tlambert03: yes, should check that it's either None or callable, but don't need to verify parameter/return types

src/app_model/registries/_keybindings_reg.py
```diff
@@ -26,12 +26,30 @@ class _RegisteredKeyBinding(NamedTuple):
...
 class KeyBindingsRegistry:
-    """Registry for keybindings."""
+    """Registry for keybindings.
+
+    Attributes
+    ----------
+    filter_keybinding : Callable[[KeyBinding], str] | None
+        Function for filtering out invalid `KeyBinding`s.
+        Callable should return an error message string if `KeyBinding` is invalid
+        or empty string if `KeyBinding` is valid.
```
tlambert03, line 36:
```suggestion
        Optional function for applying additional `KeyBinding` validation.
        Callable should accept a `KeyBinding` object and return an error message (`str`)
        if `KeyBinding` is rejected, or empty string otherwise.
```

trying to make it clearer that the KeyBinding object itself will already be valid, and that this is a app-specific "preference"

src/app_model/registries/_keybindings_reg.py
tlambert03: yep

src/app_model/registries/_keybindings_reg.py
```diff
@@ -26,12 +26,30 @@ class _RegisteredKeyBinding(NamedTuple):
...
     registered = Signal()
 
     def __init__(self) -> None:
         self._keybindings: list[_RegisteredKeyBinding] = []
+        self._filter_keybinding: Callable[[KeyBinding], str] | None = None
+
+    @property
+    def filter_keybinding(self) -> Callable[[KeyBinding], str] | None:
+        """Return the `filter_keybinding`."""
+        return self._filter_keybinding
```
tlambert03, line 48: need to cover this line in tests

tlambert03 (commented):
looks good,  but we need to catch the new possible `ValueError` inside of the for loop in `register_action_keybindings`

```python
            if d := self.register_keybinding_rule(action.id, _keyb):
                disposers.append(d)
```

otherwise one bad keybinding will ruin the whole batch (and add a test for that)

lucyleeow:
Thanks @tlambert03, I think changes are made. I will look at any doc updates once code okay.

tlambert03 (commented):
thanks @lucyleeow, perfect 👌

PR #218, opened 2024-08-14
feat: add action keybinding info over tooltip

### LINKED ISSUE #216: feat: include action current keybinding as part of its tooltip
### Description

When an action has a keybinding set it could be nice to be able to see it as part of the action's tooltip. So, for example, over napari,you can see tooltips formated in the following way:

imagen

This is part of the elements that started to be discussed over an exploration to use `QModelToolBar`s for buttons over napari. For more details on that you can see https://github.com/napari/napari/pull/7133

dalthviz:
Closes #216
Related to https://github.com/napari/napari/pull/7133

### DIFF
--- a/src/app_model/types/_keys/_standard_bindings.py
+++ b/src/app_model/types/_keys/_standard_bindings.py
@@ -70,6 +70,7 @@ class StandardKeyBinding(Enum):
     Underline = auto()
     Undo = auto()
     WhatsThis = auto()
+    OriginalSize = auto()
     ZoomIn = auto()
     ZoomOut = auto()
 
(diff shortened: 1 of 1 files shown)


src/app_model/backends/qt/_qaction.py
```diff
@@ -54,6 +56,11 @@ def _on_triggered(self, checked: bool) -> None:
         # to raise any exceptions.
         self._app.commands.execute_command(self._command_id).result()
 
+    def setToolTip(self, tooltip: str) -> None:
```
tlambert03, line 59: with this pattern, is there any way for someone to opt *out* of this behavior?  For example, if they wish to set their own tooltip (sans suffix).  What do you think about simply calling `self.setToolTip` with the text and the shortcut suffix in `__init__` (and then later calls to `setToolTip()` would override it).  ... or some other method that makes it opt-out-able

dalthviz: Oh I see, maybe I overdid a little bit here trying to implement a way to always show the keybinding 😅 Will simplify things then :+1:

Also, regardless of the simplification. could it make sense to add over the `Action` definition a field like the one for `CommandRule` definition for the icon visibility?

https://github.com/pyapp-kit/app-model/blob/4afa6f6fcf046bb7e48aa91f88f21cefc5b2352b/src/app_model/types/_command_rule.py#L61-L65

So something like `keybinding_visible_in_tooltip` and make its default value `False`?

tlambert03: i think i'd prefer not to have that sort of "view-specific" setting in the model itself.

DragaDoncila:
@tlambert03 I think your comments have been addressed if you wouldn't mind taking another look at this PR?

tlambert03 (approved):
thanks!

PR #219, opened 2024-08-15
feat: add support for '.svg' file paths as action icons

### LINKED ISSUE #217: feat/question: usage of `.svg` sources directly as `Action`s icons and how to customize `Action`'s icon color?
### Description

Currently, over napari, some `.svg` sources are used to generated themed/colored icons + QSS to use the generated files/icons over Qt elements. Could it be worthy to be able to allow to pass as an `Action`'s `icon` definition a path to a `.svg` besides `iconify`/`supertqt.fonticon` identifiers?

If that doesn't seem like the right approach, what would be the proper way to provide/set colored `Action`'s icons? I saw https://github.com/pyapp-kit/app-model/pull/132 (although seems like there only a dark and a light color could be configured?) and checking superqt `fonticon` utilities seems like you can create font plugins, right? Should napari eventually try to create a font including the `.svg` files currently being used? 🤔 

Edit: This is part of the elements that started to be discussed over an exploration to use `QModelToolBar`s for buttons over napari. For more details on that you can see https://github.com/napari/napari/pull/7133

dalthviz:
Closes #217

### DIFF
--- a/src/app_model/backends/qt/_util.py
+++ b/src/app_model/backends/qt/_util.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
+from qtpy.QtCore import QUrl
 from qtpy.QtGui import QIcon
 
 if TYPE_CHECKING:
@@ -15,7 +16,9 @@ def to_qicon(icon: Icon, theme: Literal["dark", "light"] = "dark") -> QIcon:
     from superqt import QIconifyIcon, fonticon
 
     if icn := getattr(icon, theme, ""):
-        if ":" in icn:
+        if "file:///" in icn:
+            return QIcon(QUrl(icn).toLocalFile())
+        elif ":" in icn:
             return QIconifyIcon(icn)
         else:
             return fonticon.icon(icn)


DragaDoncila:
@tlambert03 what do you think of this PR?

tlambert03:
I'm out of the country for another  few days, sorry for the delay

DragaDoncila:
> I'm out of the country for another few days, sorry for the delay

no worries at all! Just wanted to ping in case it got forgotten 😊

demo/model_app.py
```diff
@@ -231,6 +235,7 @@ class CommandId:
     ),
     types.Action(
         id="about",
+        icon=f"file://{ABOUT_ICON_PATH}",
```
dalthviz, line 238: Hi there, thank you for pushing foward this! Quick comment, I think there is a need (at least on Windows) to put three slashes here (`///`) otherwise no icon is shown:

imagen

tlambert03: i see, apologies for changing that!  I thought the third slash was an oversight (since there's also a preceding slash for posix systems in ABOUT_ICON_PATH).

tlambert03, line 238:
```suggestion
icon=f"file:///{ABOUT_ICON_PATH}",
```

dalthviz: Oh I see, maybe we should show over the example the need to handle the slashes for `file:` depending on the platform? Checking https://en.wikipedia.org/wiki/File_URI_scheme I see that on Unix the preceding slash is taken into account for the three slashes the URI has before the actual file path :thinking:

tlambert03: Based on the fact that your previous version was working fine (and tests are passing again after reverting it), it appears to be fine for posix systems to have four slashes (`file:////Users...`).  So I'm fine leaving it as is and not doing anything platform specific.  But I suppose we could add a note somewhere?  Your call, I'm fine just merging

tlambert03: Woah I said "fine" a lot in there 😅

dalthviz: Over the last commit [adffa9b](https://github.com/pyapp-kit/app-model/pull/219/commits/adffa9b330f5d3834616e2059c433f0777a5dd1e) I added some info about the possibility to use a `.svg` file path over the Action `icon` field with a note just in case. Seems like an approapiate place but let me know what you think!

tlambert03: perfect!

tlambert03:
all set for merge?

dalthviz:
Yep :+1:

PR #226, opened 2024-11-29
feat: initial exploration for keybinding source addition and inverse map for keybinding registry

### LINKED ISSUE #207: Store 'source' info in `KeyBindingRegistry`
Enable `KeyBindingRegistry` to store information on the 'source' of the keybinding e.g., whether user, extension or core set the keybinding. Sources will have priority order e.g., user > ext > core

I am not sure whether the source options are immutable, set to be user > ext > core , or if we allow the user to specify 'source' and their priority order. In the latter case we could default to user > ext > core. WDYT @tlambert03 @dragaDoncila ?

>  So get_keybinding should be a function that takes a pressed key, (potentially) narrows keybindings based on context, and then picks a command based on the prioritization of the source

@tlambert03 is this different from `KeyBindingRule`s `when` (which currently is not implemented)?

Migrated from discussion in #177

dalthviz:
Closes #207

Based on https://github.com/napari/napari/pull/6204

Currently this:

* Adds a `source` field and three prioritized sources (`USER`, `PLUGIN`, `SYSTEM`)
* Adds a new method that returns the highest priority active keybinding given the evaluated context (`get_context_prioritized_keybinding`)
* Adds an inverse map keyable by the key sequence sorting the keymap values using the source and weight (`_keymap`)
* Changes the `_keybindings` attribute to be computed from the new `_keymap` attribute
* Adds tests for the new `_RegisteredKeyBinding` comparison logic and the method to retrieve a prioritaized keybinding (`get_context_prioritized_keybinding`)


DragaDoncila:
I had a look at this and I like the implementation, and sorting first by source, then by weight. The only question that came to mind is whether we think the sources should also be app-configured or whether we think `USER > PLUGIN > SYSTEM` is generic enough. I'm also thinking that `SYSTEM` might be ambiguous as OS-defined codes, and maybe we want to change that to `APP`?

@dalthviz it would be good to update the description to add specifically what is changing namely:

- adding source field and three prioritized sources
- adding new method that returns the highest priority active keybinding given the evaluated context
- adding an inverse map keyable by the key sequence
- sorting the keymap values using the source and weight

@tlambert03 what do you think about the changes here?

src/app_model/types/_constants.py
```diff
@@ -3,6 +3,14 @@
 from enum import Enum
 
 
+class KeyBindingSource(Enum):
```
tlambert03, line 6: would be a good candidate for `IntEnum`, to gain support for rich comparisons

src/app_model/registries/_keybindings_reg.py
```diff
@@ -38,9 +64,11 @@ class KeyBindingsRegistry:
     """
 
     registered = Signal()
+    unregistered = Signal()
 
     def __init__(self) -> None:
         self._keybindings: list[_RegisteredKeyBinding] = []
+        self._keymap: dict[int, list[_RegisteredKeyBinding]] = {}
```
tlambert03, line 71:
```suggestion
        self._keymap = defaultdict[int, list[_RegisteredKeyBinding]](list)
```

and import defaultdict from collections

src/app_model/registries/_keybindings_reg.py
```diff
@@ -133,13 +162,31 @@ def register_keybinding_rule(
...
             )
             self._keybindings.append(entry)
+
+            # inverse map registry
+            kb = keybinding.to_int()
+            if kb not in self._keymap:
+                entries: list[_RegisteredKeyBinding] = []
+                self._keymap[kb] = entries
+            else:
+                entries = self._keymap[kb]
```
tlambert03, line 175: after turning `_keymap` into a defaultdict, this becomes:
```suggestion
            entries = self._keymap[keybinding.to_int()]
```

src/app_model/registries/_keybindings_reg.py
```diff
@@ -23,8 +29,28 @@ class _RegisteredKeyBinding(NamedTuple):
...
+        if not isinstance(other, _RegisteredKeyBinding):
+            return NotImplemented
+        return self.source.value < other.source.value or (
+            self.source == other.source and self.weight < other.weight
+        )
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, _RegisteredKeyBinding):
+            return NotImplemented
+        return self.source.value == other.source.value and self.weight == other.weight
```
tlambert03, line 52: after making `KeyBindingSource` an `IntEnum`:

```suggestion
    def __gt__(self, other: object) -> bool:
        if not isinstance(other, _RegisteredKeyBinding):
            return NotImplemented
        return (self.source, self.weight) > (other.source, other.weight)

    def __lt__(self, other: object) -> bool:
        if not isinstance(other, _RegisteredKeyBinding):
            return NotImplemented
        return (self.source, self.weight) < (other.source, other.weight)

    def __eq__(self, other: object) -> bool:
        if not isinstance(other, _RegisteredKeyBinding):
            return NotImplemented
        return (self.source, self.weight) == (other.source, other.weight)
```

src/app_model/registries/_keybindings_reg.py
tlambert03: Also, I don't think we need to maintain *both* `_keybindings` and `_keymap` do we?  I believe you could just delete `self._keybindings` here, get rid of the `_keybindings.append` and `_keybindings.remove` logic in `register...`, and replace it with a property that just iterates through all the lists in `_keymap`:

```python
    @property
    def _keybindings(self) -> Iterable[_RegisteredKeyBinding]:
        return (entry for entries in self._keymap.values() for entry in entries)
```

then you'll also need to remove `len(self._keybindings)` from `__repr__` ... and could use `len(self)` instead with:

```python
    def __len__(self) -> int:
        return sum(len(entries) for entries in self._keymap.values())
```

src/app_model/registries/_keybindings_reg.py
```diff
@@ -150,9 +197,45 @@ def __repr__(self) -> str:
...
+        registered_keybindings_list = self._keymap[key].copy()
+        registered_keybindings_list.reverse()
+        return next(
+            (
+                entry
+                for entry in registered_keybindings_list
+                if entry.when is None or entry.when.eval(context)
+            ),
+            None,
         )
```
tlambert03, line 241:
```suggestion
        if key in self._keymap:
            for entry in reversed(self._keymap[key]):
                if entry.when is None or entry.when.eval(context):
                    return entry
        return None
```

tlambert03 (commented):
looks good!  made a number of suggestions for cleanup of logic and syntax

dalthviz:
> The only question that came to mind is whether we think the sources should also be app-configured or whether we think USER > PLUGIN > SYSTEM is generic enough. I'm also thinking that SYSTEM might be ambiguous as OS-defined codes, and maybe we want to change that to APP?

Over https://github.com/napari/napari/pull/6204 I think the value is named `CORE`. Other name could be `BASE`. Not totally sure what could be the best option to be honest 😅 but, in case having a fixed set of values is the way to go, while working on this I would say I have seen the following options:

* `USER` > `PLUGIN` > `CORE`
* `USER` > `EXT` > `CORE`
* `USER` > `PLUGIN` > `SYSTEM`
* `USER` > `PLUGIN` > `APP`
* `USER` > `PLUGIN` > `BASE`

Let me know what could be the prefered combination or if providing an option to use a custom set of values/enum makes sense!

tlambert03:
> Let me know what could be the prefered combination or if providing an option to use a custom set of values/enum makes sense!

I'm happy to defer to the group opinion/decision here.  I like `APP` or `CORE` or `BASE` over system.  And don't have strong opinion on `EXT` v `PLUGIN`.

thanks for all this @dalthviz

dalthviz:
> I'm happy to defer to the group opinion/decision here. I like APP or CORE or BASE over system.

For the moment then I will follow the naming suggestion done by Draga (last commit changes `SYSTEM` for `APP`)

Thank you for the reviews @tlambert03 @DragaDoncila and let me know if something else is needed here!

lucyleeow:
Late to the party but this is great @dalthviz. Great choices, will be nice to use for keybinding implementation in napari.

PR #230, opened 2024-12-15
feat: add KeyCombo.Ctrl and KeybindingRule.to_keybinding

### LINKED ISSUE #229: Question: how to specify KeyMod of Control for all platforms?
KeyMod.CtrlCmd results in Ctrl on windows (and linux?) but Cmd on macOS
KeyMod.WinCtrl says it does Meta on Windows and Ctrl on macOS

Is there a way to specify a Control-Foo binding that uses Control on all platforms?
KeyMod.Ctrl does't exist, but there is KeyCode.Ctrl

Based on 
https://github.com/pyapp-kit/app-model/blob/a0d95b19c0e8d61ccab6bea69d1adace910d9157/src/app_model/backends/qt/_qkeymap.py#L220-L238

it looks like it wouldn't be easy to have two different KeyMod map to the same QCTRL...

tlambert03:
closes #229

This PR adds two more KeyMods:  `KeyMod.Ctrl` and `KeyMod.Meta`

- CtrlCmd:   command on a mac, control on windows
- Shift:   shift key
- Alt:  alt/option key
- WinCtrl:  meta key on windows, control key on mac
- **Ctrl**:  control key, regardless of OS (NEW)
- **Meta**: command key on a mac, meta key on windows (NEW)

most of the time, you will want to use `CtrlCmd` and `WinCtrl`, since they do what is usually "expected" when the keycode is converted to a Keybinding object.   But in cases where you just always want the control key, or meta key, regardless of platform, the new codes can be used.  Note: it *can* be confusing if you combine `Ctrl` with `CtrlCmd` ... (it would be just ctrl on windows, but cmd+ctrl on macos) so it's definitely not recommended.  

It also adds some conveniences for debugging.  It adds a public `KeyBindingRule.to_keybinding` method that takes in an optional os.  so one can easily test how these things are resolved, on any platform

```python
from app_model.types import KeyBindingRule, KeyCode, KeyMod, OperatingSystem

rule = KeyBindingRule(primary=KeyMod.CtrlCmd | KeyCode.KeyX)
os = OperatingSystem.WINDOWS
print(rule.to_keybinding(os=os).to_text(os=os))
```

### DIFF
--- a/src/app_model/registries/_keybindings_reg.py
+++ b/src/app_model/registries/_keybindings_reg.py
@@ -6,8 +6,6 @@ from typing import TYPE_CHECKING, Callable, NamedTuple
 
 from psygnal import Signal
 
-from app_model.types import KeyBinding
-
 if TYPE_CHECKING:
     from collections.abc import Iterable, Iterator, Mapping
     from typing import TypeVar
@@ -16,6 +14,7 @@ if TYPE_CHECKING:
     from app_model.types import (
         Action,
         DisposeCallable,
+        KeyBinding,
         KeyBindingRule,
         KeyBindingSource,
     )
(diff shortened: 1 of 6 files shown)


tlambert03:
@dalthviz, would appreciate your review.  (also, added you to collaborators here)

dalthviz (commented):
Thanks @tlambert03 for the invitation! I think the only suggestion from my side would be to somehow see a way to add the info you put over the PR description in the documentation. Maybe adding some of that info over the `KeyMod` class docstring could be a way of achieving that? Or maybe is there a way for the docs to show a link to go to the source so people can check the inline comments on the enum values to better graps who the enum values interact with the different platforms/operative systems? 🤔

dalthviz (approved):
Hi there and Happy New Year! Just in case, gave this another check and related with the two additional `KeyMod` values and their interaction with other values (i.e `Ctrl` + `CtrlCmd` ), played a little bit changing the `KeyMod` enum docstrings and doing something like:

```python
class KeyMod(IntFlag):
    """
    A Flag indicating keyboard modifiers.

    Note: :warning: Combining `Ctrl` with `CtrlCmd` is not recommended. That combination of
    keyboard modifiers behaves in a specific way depending on the operative system meaning
    `Ctrl` on Windows but ending up as `Cmd + Ctrl` on macOS. 
    """

    NONE = 0
    """None key"""

    CtrlCmd = 1 << 11
    """command on a mac, control on windows"""

    Shift = 1 << 10
    """shift key"""

    Alt = 1 << 9
    """alt/option key"""

    WinCtrl = 1 << 8
    """meta key on windows, control key on mac"""

    Ctrl = 1 << 12
    """control key, regardless of OS"""

    Meta = 1 << 13
    """command key on a mac, meta key on windows"""
```

Can end up making the docs look something like:

image

Also, for comparison this is the current entry for the enum over the docs:

image

https://app-model.readthedocs.io/en/latest/reference/types/#app_model.types.KeyMod

Anyhow leaving this approved :+1:

psobolewskiPhD:
This looks really great @tlambert03 
Thanks!

PR #241, opened 2025-04-28
feat: allow explicit variable bound to be passed to Name

tlambert03:
this allows for directly providing the type to which a Name expression evaluates to:

```python
from app_model.expressions import Name

is_running = Name("is_running", bound=bool)
reveal_type(is_running.eval())  # Revealed type is "builtins.bool"
```

also updates pre-commit

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,13 +5,13 @@ ci:
 
 repos:
   - repo: https://github.com/crate-ci/typos
-    rev: v1
+    rev: v1.31.1
     hooks:
       - id: typos
         args: []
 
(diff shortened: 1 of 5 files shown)


tlambert03:
@jni or @Czaki  ... do you guys know what might be going on with the failed napari test here? https://github.com/pyapp-kit/app-model/actions/runs/14710881217/job/41282696495?pr=241

jni:
It's napari/napari-plugin-manager#151, sorry about that! Looks like a fix is incoming. (napari/napari-plugin-manager#152) 🙏

jni:
(ie safe to ignore the failure, will be fixed separately)

PR #246, opened 2025-06-18
fix: Use QApplication instance as a parent of global actions

### LINKED ISSUE #245: Singleton of Actions do not work with multiple windows of same application
In napari we got from users information that menus disappear if user close latest created window.  

https://github.com/napari/napari/issues/7588
https://github.com/napari/napari/issues/8024

My investigation shows that it looks to be caused to a singleton nature of Actions in app-model. 

https://github.com/pyapp-kit/app-model/blob/06c1bc449d9bbd90438face1530206b1859460bd/src/app_model/backends/qt/_qaction.py#L178-L182 

As the latest created window is set to be a parent of actions, then, on close of this window, Qt destroys of actions, so the actions effectively disappear from all windows. 

What is the motivation to use singleton of `QMenuItemAction`, when `QModelSubmenu` are not singletons? 

If it is for update, shouldn't we store them per `QMainWindow` instance?

Czaki:
Closes #245

I tested it locally, do not have an idea how to test it on CI.

### DIFF
--- a/src/app_model/backends/qt/_qmenu.py
+++ b/src/app_model/backends/qt/_qmenu.py
@@ -4,7 +4,7 @@ import contextlib
 from collections.abc import Collection, Iterable, Mapping, Sequence
 from typing import TYPE_CHECKING, cast
 
-from qtpy.QtWidgets import QMenu, QMenuBar, QToolBar
+from qtpy.QtWidgets import QApplication, QMenu, QMenuBar, QToolBar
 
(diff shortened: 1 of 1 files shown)


tlambert03:
slight preference for this over #247, since A) it still gives the action a parent, which is nice for app-level cleanup, and B) I wonder whether tying the actions to a Menu or Toolbar is too specific?

PR #275, opened 2026-02-19
fix: themed color for icons

brisvag:
When icons are provided via fonticons or iconify, theme is currently disergarded. Defaulting to black/white should be at least a tiny bit nicer than just black!

### DIFF
--- a/src/app_model/backends/qt/_util.py
+++ b/src/app_model/backends/qt/_util.py
@@ -15,7 +15,7 @@ def to_qicon(icon: Icon, theme: Literal["dark", "light"] = "dark") -> QIcon:
     """Create QIcon from Icon."""
     from superqt import QIconifyIcon, fonticon
 
-    color = 'white' if theme == 'dark' else 'black'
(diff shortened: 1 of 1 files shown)


brisvag:
Mmm, just realized the theme is not passed through at all :thinking:

tlambert03:
see also https://github.com/pyapp-kit/app-model/pull/132 which was trying to improve icon color support.   let me know what overlaps

brisvag:
Ah, silly me, I didn't check before opening. It's definitely covered there, sorry for the noise!

tlambert03:
well, it's not noise!  Cause that PR is two years old with no activity :joy:  it would be awesome, if this is something you're wanting to see changed, for you to take it over the finish line.  i can't remember anymore any of the details... but feel free to go any direction you want with it

tlambert03:
it's also fine if the scope of that PR makes too big what was just a simple fix and you want this in instead

brisvag:
Yeah I might try to take it up soon, we'll see if I can figure it out :)

PR #278, opened 2026-04-10
WIP: KeyBinding: Flexible equality

gselzer:
Hi!

I was experimenting with this repo over in scenex for its key bindings model, and I was finding it tedious to check key binding equality. For example, if I want to know whether the left arrow was pressed, it'd be nice to say "kb == KeyCode.LeftArrow`, but that does not work as is. Was wondering whether a syntax like the following would be problematic?

Of course, if we like more "magical" equality checks, there are probably more things to do. Other equality checks, hashing checks, etc.


gselzer:
Ooh, interesting, now I'm reading https://github.com/pyapp-kit/app-model/issues/108. Maybe this wasn't the best approach, let me take another stab when I next get a chance 😅

PR #285, opened 2026-05-11
tests: Use dependency-group for napari testing

TimMonko:
napari/napari#8704 is aiming to drop `optional-dependencies` this uses the `dependency-group` definition from https://github.com/pyapp-kit/workflows/blob/main/.github/workflows/test-dependents.yml

🍻

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,7 +141,7 @@ jobs:
     secrets: inherit
 
   test_napari:
-    uses: pyapp-kit/workflows/.github/workflows/test-dependents.yml@v2
+    uses: pyapp-kit/workflows/.github/workflows/test-dependents.yml@65abcc0b091a0d45dbe3c589266160cc9b62f8dc
     with:
       dependency-repo: napari/napari
       dependency-group: "testing"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,7 +141,7 @@ jobs:
     secrets: inherit
 
   test_napari:
-    uses: pyapp-kit/workflows/.github/workflows/test-dependents.yml@65abcc0b091a0d45dbe3c589266160cc9b62f8dc
+    uses: pyapp-kit/workflows/.github/workflows/test-dependents.yml@main
     with:
       dependency-repo: napari/napari
       dependency-group: "testing"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -141,10 +141,10 @@ jobs:
     secrets: inherit
 
   test_napari:
-    uses: pyapp-kit/workflows/.github/workflows/test-dependents.yml@v2
+    uses: pyapp-kit/workflows/.github/workflows/test-dependents.yml@65abcc0b091a0d45dbe3c589266160cc9b62f8dc
```
TimMonko: It seems this commit hasn't been released (that contains dependency-groups), so let me know how you want to handle this :)

tlambert03: same comment as https://github.com/pyapp-kit/superqt/pull/344#issuecomment-4424356643

just use `@main`

TimMonko: thanks @tlambert03 wasn't sure how you wanted to handle it :)

TimMonko:
```suggestion
    uses: pyapp-kit/workflows/.github/workflows/test-dependents.yml@main
```

TimMonko:
just a bit of feedback because I keep screwing up, the whole Semantic PR fail just links to a repo and eventually I figure out it wants conventional commits but I keep forgetting to use conventional commits because nothing blocks me 🤷

tlambert03:
it's fine.  in these cases i/we just change the title of the PR, and it's fixed.  nothing the contributor needs to do (but you can also change the title to match CC if you want to make it all go green).  (you *don't* need to change the commit itself)

TimMonko:
> it's fine. in these cases i/we just change the title of the PR, and it's fixed. nothing the contributor needs to do (but you can also change the title to match CC if you want to make it all go green). (you _don't_ need to change the commit itself)

ohh interesting. Thanks :)

tlambert03:
yeah; The "commit" in question here is the final squash/merge commit... not every commit that you do as a contributor.  that would be draconian.  as is, it's just useful for that final merge to have CC, assists in autogenerating changelogs

TimMonko:
cc @Aniketsy since we were working on all these things together

